PR #10, opened 2021-03-29
Added an `html()` method to the `Dashboard` class

sllynn:
As per the title: added an `html()` method to the `Dashboard` class to support displaying dashboards in other notebook platforms e.g. Databricks.


emeli-dral:
Thanks for the commit, we add your changes to the project! The only point I want to draw your attention to is embedding the generation of our html in other modules (especially at this stage of the project😅). In this case, some collisions are possible, styles or some sections of the report may not be displayed correctly.

PR #54, opened 2021-09-14
fix exception: "local variable confidence referenced before assignment"

RadionBik:


RadionBik:
This fixes an exception I noticed when updated to the latest version (0.1.23.dev0)

emeli-dral:
Hi @RadionBik ,

makes sense, thank you a lot for the fix!

RadionBik:
Hi! Could you create a new tag with this (or latest) fix, please? So others could use a package installed via `pip install evidently` without bugs :)

PR #70, opened 2021-10-20
PR to change zero_trace on graph

Pakard1:
PR to change name of zero_trace to Reference and showlegend True in all graphs to be possible to filter out this trace.

Reg_pred_and_actual_in_time_widget.py zero_trace has showlegend True but by default it is not selected


emeli-dral:
Hey!
Unfortunately,  the library version has gone far ahead, and we do not accept the pull request.
Although, the idea of this change is good: we would like to leave the green line on the chart by default but allow turning it off.

I do apologize we lost this pull request. We will do our best to avoid such situations in the future and no longer lose useful changes.

PR #82, opened 2021-11-25
refactor(utils): Refactoring of the NumpyEncoder

burkovae:
Here a very small refactoring proposal, that actually fixes some (i guess rare) parsing errors in the `NumpyEncoder`.

* 🐞 Some `numpy` types could not be parsed with a previous version
* 🐞 If a non-numpy type was given to the encode, it would result in a `TypeError` as per https://docs.python.org/3/library/json.html because of `json.JSONEncoder.default(self, obj)` call
* It seems the `NumpyEncoder` was copypasta in some example file. Not anymore.
* Also the refactored code contains less lines of code and delegates typing to `numpy` itself
* Please not, that `np.void(3)` would create an empty byte buffer of length 3. It sills convert to `None`. But keep in mind that this conversion is arbitrary and may cause error in a far future.


Liraim:
Hi,

Thanks for contribution. There are some really useful fixes.

About `np.void` - currently there no "good" solution for this conversion, so it is there only for preventing it from breaking serialization to JSON.

And I have only one required change to this PR about tests: you need to move tests to folder where source file is located (to `evidently/utils/test_utils.py`) and revert changes to CI scripts, so this PR wont change previous behaviour.

About changing location of tests: currently we place tests nearby of tested files, and we are in discussion about best approach for us.

burkovae:
Thank you for your feedback. The change to CI does not change previous behaviour - I wonder what you mean by that. Anyhow, I reverted the change. Please consider using default behaviour as a conventional approach - adding parameters just raises questions. Imho simply `unittest discover` just does the job well already.

As for discussion about testing framework.
Imho, you should consider the following:
* having a dedicated directory is a suggested way [nowadays](https://docs.python-guide.org/writing/structure/#sample-repository):
  * See for example [pytest docs about tests outside applichation](https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-outside-application-code)
   > Putting tests into an extra directory outside your actual application code might be useful if you have many functional tests or for other reasons want to keep tests separate from actual application code (**often a good idea**):
  * But of course there are good contrary examples in pandas, pytorch, tensorflow, sklearn etc., i.e. [test as part of application code](https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-as-part-of-application-code)
  > Inlining test directories into your application package is useful if you have direct relation between tests and application modules and want to distribute them along with your application
   * Still, even in these example, there is a dedicated test directory with all the tests. Tests *files* inside a submodule directory is very uncommon.
* I'm personally a fan of  tests outside applichation:
   * it enforces a good TDD practice, since you can immediatly see what tests are responsible for what.
   * I don't think including tests is practical for the end user. I don't see a point in `import evidently.test_app`.
   * If you pack tests in your application code, they too will be style-checked - However, good functional and behavioral tests often do not comply with predefined styles in my experience. There way too many issues which are conter-productive when you want to make writing tests easier.
   
As for test framework, I would encourage you to checkout [pytest](https://docs.pytest.org/en/6.2.x/getting-started.html#create-your-first-test). It makes writing tests way easier and promotes good programming practice (as opposed to unittest, _imho_).

When you reached your decision, consider putting such information into your contributing guide for the future.

burkovae:
And here is also an [article that really make a good case for pytest](https://blog.daftcode.pl/the-cleaning-hand-of-pytest-28f434f4b684). I hope this information will help you to reach a conclustion.

Liraim:
Thanks for sharing information about pytest, it definitely will help to make a decision faster!

PR #90, opened 2021-12-04
Use pytest instead of unittest

Tapot:
Use Pytest instead of unittest python module
- add pytest==6.2.5 to dev requirements
- add pytest settings to setup.cfg - just specify main python code place
- update CONTRIBUTING.md about tests run + some description typo fixes in the section
- update github actions for running tests
- update tests syntax from uniitests to more-pytest style - use asserts instead of self.assert methods


burkovae:
If there is a decision to move to `pytest`, please consider moving tests into a separate directory as well. (As for the test directory structure, there're strategies as well)

Tapot:
> If there is a decision to move to `pytest`, please consider moving tests into a separate directory as well. (As for the test directory structure, there're strategies as well)

I think that it is a better way too, it can be the second step in s separate PR - for better diff readability

Let's wait for a maintainer opinion

evidently/analyzers/test_utils.py
```diff
@@ -1,13 +1,12 @@
...
-import unittest
 
 import pandas
 
 from evidently.analyzers.utils import process_columns
 from evidently.pipeline.column_mapping import ColumnMapping
 
 
-class TestUtils(unittest.TestCase):
+class TestUtils:
```
Liraim, line 9: class can be removed, and all methods can be made just regular top level functions.

evidently/utils/test_utils.py
```diff
@@ -1,38 +1,40 @@
...
-class TestNumpyEncoder(unittest.TestCase):
+@pytest.fixture
+def numpy_encoder():
+    return NumpyEncoder()
 
-    def setUp(self) -> None:
-        self.encoder = NumpyEncoder()
 
-    def test_int_convert(self):
+class TestNumpyEncoder:
```
Liraim, line 13: same here, class can be removed.

Liraim (changes requested):

evidently/analyzers/test_utils.py
Tapot: I've moved the tests to functions, classes are removed for already moved to pyrest test

Tapot:
I'm going to move new tests to pytest in the PR and I can move all tests to one folder. Or maybe moving tests to a separate folder will be better in another PR, @Liraim what do you think?

Liraim:
> I'm going to move new tests to pytest in the PR and I can move all tests to one folder. Or maybe moving tests to a separate folder will be better in another PR, @Liraim what do you think?

I will move tests to separate folder in another PR. So nothing is required here.

Liraim (approved):
Looks good.

Liraim (changes requested):
Need fix for linter.

Liraim (approved):
Now Looks good.

PR #91, opened 2021-12-06
Add some basic tests to `Dashboard`, `Profile`, and `ProbClassificationPerformanceAnalyzer`

### LINKED ISSUE #89: Input for ProbClassificationPerformanceAnalyzer
I posted same question at discord first, but this maybe a more appropriate place. We'll see ;-)

I'm writing some tests for dashboards and profiles but stumbled upon a probable bug in `ProbClassificationPerformanceAnalyzer`.
https://github.com/burkovae/evidently/blob/ab-fix-classification-bugs/evidently/analyzers/test_classification_analyzers.py

Since I'm not sure, how a correct input would look like, I hope that someone can post ist here. I will be happy with a very basic data for an analysis of a single dataframe. Hence, I hope to get some correct input, so that I can narrow down what's happening and what components fails.

Especially I would like know, what the `ProbClassificationPerformanceAnalyzer` computes exactly (@emeli-dral it seems you wrote this piece). I would be happy with few inputs and expected outputs. 

If you already know the fix to

```
Error
Traceback (most recent call last):
  File "...\evidently\evidently\analyzers\test_classification_analyzers.py", line 21, in test_single_dataset_with_two_classes
    blubb = analyzer.calculate(df, None, ColumnMapping())
  File "...\evidently\analyzers\prob_classification_performance_analyzer.py", line 32, in calculate
    prediction_labels = [prediction_column[x] for x in prediction_ids]
TypeError: 'numpy.intc' object is not iterable
```

then I urge you to write the tests with inputs and expected outputs first, please. It helps down the line and improves robustness and safeguard against unwanted changes.

If you can point me in the right direction, I will fix any issues along the road. I just need to know what you compute specifically with some explicit examples.

burkovae:
Before refactoring working code, it is a good idea to write tests that model the current behavior since small changes may break the components.

These are a few tests for code parts, that I tried to understand. The added tests serve as a starting point for refactoring. I hope the tests will evolve with time and become more concise and clean.

Bugs discovered along the way:

* 🐞 In README.md for `RegressionPerformanceProfileSection` to work one need to provide a correct column mapping for `prediction` because otherwise the result will be empty
  * Fix: either add additional `prediction` column in README.md
  * e.g. `self.iris_frame['prediction'] = iris.target[::-1]`
* 🐞 ~~some weird stuff is happening in the `ProbClassificationPerformanceAnalyzer`. Because `prediction_ids` is just an integer, we cannot iterate for prediction labels.~~ The input should be correct one, of course 🤐
* 🐞 some other weird stuff is happening in the `ProbClassificationPerformanceAnalyzer`
  * see `FIXME` tags in [test file](https://github.com/burkovae/evidently/pull/3/files#diff-ab345753bb56c93445789566698667f01206d5c07117e7d2f94b07bbcc0f649aR263-R265)
 
* 🐱‍👤make `column_mapping` an optional argument (with reasonable defaults) in `Profile` so that it is compatible with previous behavior.

closes #89


Liraim (approved):
Looks good. But try avoid TODO comments in next commits.

PR #99, opened 2021-12-12
Reorganize code

Liraim:


svenaoki (commented):
pip install on Windows wont move the Tab classes into evidently.dashboard.tabs but they remain under the evidently.tabs folder

Liraim:
> pip install on Windows wont move the Tab classes into evidently.dashboard.tabs but they remain under the evidently.tabs folder

Hi, this change isn't in the version on pypi yet.

PR #110, opened 2021-12-26
Add typed results for analyzers

Tapot:
- add DataDriftAnalyzerResults for DataDriftAnalyzer
- move data drift analyser results to DataDriftAnalyzerResults instance, not dict (TODO: move metrics from dict to a dataclass object)
- simplify metrics names iteration (TODO: add some lazy calculations optimisations)
- add a test for DataDriftAnalyzerResults serialization to dict

aux changes:
- add mypy and pytest caches to git ignore settings

### DIFF
--- a/evidently/analyzers/data_drift_analyzer.py
+++ b/evidently/analyzers/data_drift_analyzer.py
@@ -1,15 +1,17 @@
 #!/usr/bin/env python
 # coding: utf-8
 import collections
+from typing import Any, Dict, List
 
 import pandas as pd
 import numpy as np
+from dataclasses import dataclass, field
 
 from evidently import ColumnMapping
 from evidently.analyzers.base_analyzer import Analyzer
 from evidently.options import DataDriftOptions
 from evidently.analyzers.stattests import chi_stat_test, ks_stat_test, z_stat_test
-from evidently.analyzers.utils import process_columns
+from evidently.analyzers.utils import process_columns, DatasetUtilityColumns
 
 
 def dataset_drift_evaluation(p_values, drift_share=0.5):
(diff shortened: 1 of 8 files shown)


evidently/analyzers/data_drift_analyzer.py
```diff
@@ -22,19 +24,55 @@ def dataset_drift_evaluation(p_values, drift_share=0.5):
...
 
 
+@dataclass
+class DataDriftAnalyzerResults:
+    utility_columns: DatasetUtilityColumns
+    cat_feature_names: List[str]
+    num_feature_names: List[str]
+    target_names: List[str]
+    options: DataDriftOptions
+    metrics: dict = field(default_factory=dict)
```
Liraim, line 34: I think, `metrics` better to convert to object (probably with some redesigning in structure).

Tapot: done, for now just for metrics in data drift analyzer

evidently/analyzers/data_drift_analyzer.py
```diff
@@ -22,28 +24,94 @@ def dataset_drift_evaluation(p_values, drift_share=0.5):
...
+
+    def get_all_features_list(self) -> List[str]:
+        """List all features names"""
+        return self.cat_feature_names + self.num_feature_names
+
+
 class DataDriftAnalyzer(Analyzer):
-    def calculate(self, reference_data: pd.DataFrame, current_data: pd.DataFrame, column_mapping: ColumnMapping):
+    @staticmethod
+    def get_data_drift_results(analyzer_results) -> DataDriftAnalyzerResults:
```
Liraim, line 92: I think here would be better to rename it so it won't repeat itself.
```suggestion
    def get_results(analyzer_results) -> DataDriftAnalyzerResults:
```
Is there a way to place such method to base Analyzer? So it wont be necessary to write it individually for each Analyzer.

evidently/analyzers/data_drift_analyzer.py
```diff
@@ -22,28 +24,94 @@ def dataset_drift_evaluation(p_values, drift_share=0.5):
...
 
 
+@dataclass
+class DataDriftAnalyzerFeatureMetrics:
+    current_small_hist: list
+    ref_small_hist: list
+    feature_type: str
+    p_value: float
+
+    def as_dict(self) -> Dict[str, Any]:
```
Liraim, line 34: As far as I can see `as_dict` is only used in `profile_section` (as intended, as it should return JSON as result).

Probably better to move these methods inside `profile_section` to place where it used (and purpose of this methods).

evidently/analyzers/test_data_drift_analyzer.py
```diff
@@ -0,0 +1,70 @@
...
+
+def _check_feature_metrics(feature_metric: dict, feature_type: str):
+    assert 'current_small_hist' in feature_metric
+    assert 'feature_type' in feature_metric
+    assert feature_metric['feature_type'] == feature_type
+    assert 'p_value' in feature_metric
+    assert 'ref_small_hist' in feature_metric
+
+
+def test_data_drift_analyzer_as_dict_format(analyzer: DataDriftAnalyzer) -> None:
```
Liraim, line 28: As mentioned in the `as_dict` comment: these tests should be moved to `profile_section`.

evidently/analyzers/data_drift_analyzer.py
Tapot: > I think here would be better to rename it so it won't repeat itself.
Yes, my bad,  just some overthinking, fixed

evidently/analyzers/data_drift_analyzer.py
Tapot: done - to-dict serialisation logic was moved to `profile_section` to DataDriftProfileSection

also I moved the test for dict result structure and leave separate test for DataDriftAnalyzer

evidently/analyzers/test_data_drift_analyzer.py
Tapot: done

evidently/analyzers/data_drift_analyzer.py
Tapot: About moving the method to base Analyser - need to investigate ways for types calculations with other analyzators

PR #112, opened 2021-12-29
Allow usage of Dashboard Tabs as Profile Sections

Liraim:


emeli-dral:
Decided to close it for now after an internal discussion.

PR #113, opened 2022-01-03
Fix categorical features support in Data Drift.

Liraim:


Liraim:
Fixed in another PR #141

PR #118, opened 2022-01-17
update example links to colab

elenasamuylova:
deleted "auth" at the end of the links


emeli-dral:
Thank you! 
These links are already updated, but we are looking forward to readme update 😉

PR #122, opened 2022-01-24
Quality metrics options

0lgaF:
Add quality metrics options:
conf_interval_n_sigmas
classification_threshold
cut_quantile


src/evidently/options/quality_metrics.py
```diff
@@ -0,0 +1,27 @@
...
+    cut_quantile: Union[None, Tuple[str, float], Dict[str, Tuple[str, float]]] = None
+
+    def as_dict(self):
+        return {
+            "conf_interval_n_sigmas": self.conf_interval_n_sigmas,
+            "classification_threshold": self.classification_threshold,
+            "cut_quantile": self.cut_quantile
+        }
+
+    def get_cut_quantile(self, feature_name: str) -> Tuple[str, float]:
```
Tapot, line 21: for mypy fix - set `def get_cut_quantile(self, feature_name: str) -> Optional[Tuple[str, float]]:`

src/evidently/analyzers/classification_performance_analyzer.py
```diff
@@ -31,12 +31,18 @@ def calculate(self,
 
             # calculate quality metrics
             accuracy_score = metrics.accuracy_score(reference_data[target_column], reference_data[prediction_column])
-            avg_precision = metrics.precision_score(reference_data[target_column], reference_data[prediction_column],
-                                                    average='macro')
-            avg_recall = metrics.recall_score(reference_data[target_column], reference_data[prediction_column],
-                                              average='macro')
-            avg_f1 = metrics.f1_score(reference_data[target_column], reference_data[prediction_column],
-                                      average='macro')
+            if len(prediction_column) > 2:
```
Tapot, line 34: maybe will be better to set average='binary' or average='macro' depending on len(prediction_column) and set it explicitly in the same code for both cases?

src/evidently/analyzers/classification_performance_analyzer.py
```diff
@@ -67,12 +73,18 @@ def calculate(self,
                 result['metrics']['current'] = {}
 
                 accuracy_score = metrics.accuracy_score(current_data[target_column], current_data[prediction_column])
-                avg_precision = metrics.precision_score(current_data[target_column], current_data[prediction_column],
-                                                        average='macro')
-                avg_recall = metrics.recall_score(current_data[target_column], current_data[prediction_column],
-                                                  average='macro')
-                avg_f1 = metrics.f1_score(current_data[target_column], current_data[prediction_column],
-                                          average='macro')
+                if len(prediction_column) > 2:
```
Tapot, line 76: the same as above, but for len(prediction_column)

PR #125, opened 2022-01-28
Bug fixed of a code example in readme.md

caron14:
Modify the code example in readme.md.
You might have forgotten to comment out.


Liraim:
Hi @caron14,
Yes, it seems so.
Thanks for your fix.

PR #148, opened 2022-02-09
add metrics by classes to ClassificationPerformanceMonitor

Tapot:
- add class metrics to `ClassificationPerformanceMonitor` + a separate test for it's output
- add confusion metrics by classes to `ClassificationPerformanceAnalyzer`


src/evidently/model_monitoring/monitors/classification_performance.py
```diff
@@ -33,12 +49,47 @@ def _yield_quality(
             metrics.f1, dict(dataset=dataset, metric='f1')
         )
 
+        metrics_frame = pd.DataFrame(metrics.metrics_matrix)
```
Liraim, line 52: What is a reason to convert `dict` to `pandas.DataFrame` here?

Tapot: do the same thing as in widgets implementation, I've decided that it will be better to change the same or move to analyzer and change the implementation after adding some tests

Tapot: fixed - made without pandas, just with the data from the dict

PR #150, opened 2022-02-11
add tests and some fixes for ClassificationPerformanceAnalyzer

Tapot:
`ClassificationPerformanceAnalyzer` fixes:
- calculate quality metrics once instead of twice
- fix labels calculation for confusion matrix in case we have different classes in target and prediction
- add more ClassificationPerformanceAnalyzer tests
- rename `PerformanceMetrics` to `ClassificationPerformanceMetrics`

also, raise `ValueError` from `process_columns` if we sent a None instead of a value


tests/analyzers/test_classification_performance_analyzer.py
```diff
@@ -1,41 +1,268 @@
...
+        ),
+        # no data mapping but target and prediction columns have not default names
+        (
+            pd.DataFrame({'another_target': [0, 1, 1, 0, 1], 'another_prediction': [1, 1, 0, 0, 1]}),
+            pd.DataFrame({'another_target': [0, 1, 1, 0, 1], 'another_prediction': [1, 1, 0, 0, 1]}),
+            column_mapping.ColumnMapping(),
+        ),
+    )
+)
+def test_incorrect_data_mapping_in_datasets(
```
Liraim, line 256: Probably, it should be some kind of error.

Tapot: Should I change to ValueError raising?

Liraim: I think we will address this later. First, we need to figure out how we should handle such cases.

Liraim (approved):

PR #177, opened 2022-03-09
Correlation and summary widgets

0lgaF:
- Add data summary widget
- Add correlation widget


Liraim (changes requested):
Also need to fix linter checks.

Liraim (approved):

PR #192, opened 2022-03-31
Improve Stat Tests functionality

Liraim:
- Add decorator for defining and registration of stattests
- Stattests can be used by associated name
- Custom function still can be used as stattest.

### DIFF
--- a/src/evidently/analyzers/stattests/registry.py
+++ b/src/evidently/analyzers/stattests/registry.py
@@ -1,37 +1,59 @@
 from typing import List, Union, Callable, Tuple, Dict
 
-REGISTERED_STAT_TEST: Dict[str, Dict[str, Callable]] = {}
+import dataclasses
+
+import pandas as pd
 
 
-def stattest(name: str, allowed_feature_types: List[str]):
-    def wrapper(func):
-        REGISTERED_STAT_TEST[name] = {feature_type: func for feature_type in allowed_feature_types}
-        return func
+@dataclasses.dataclass
+class StatTest:
+    name: str
+    display_name: str
+    func: Callable[[pd.Series, pd.Series, float], Tuple[float, bool]]
+    allowed_feature_types: List[str]
 
-    return wrapper
+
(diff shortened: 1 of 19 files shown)


src/evidently/analyzers/stattests/registry.py
```diff
@@ -0,0 +1,37 @@
+from typing import List, Union, Callable, Tuple, Dict
+
+REGISTERED_STAT_TEST: Dict[str, Dict[str, Callable]] = {}
+
+
+def stattest(name: str, allowed_feature_types: List[str]):
+    def wrapper(func):
+        REGISTERED_STAT_TEST[name] = {feature_type: func for feature_type in allowed_feature_types}
```
Tapot, line 8: Maybe we want to set default name as `func.__name__` ?

src/evidently/analyzers/stattests/registry.py
```diff
@@ -0,0 +1,37 @@
...
+        return func
+
+    return wrapper
+
+
+def get_stattest(stattest_func: Union[str, Callable], feature_type: str) -> Tuple[str, Callable]:
```
Tapot, line 14: Maybe use `stattest` in packages/modules naming, and in functions use one option - `stat_test` ?

tests/analyzers/stattests/test_registry.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+
+from evidently.analyzers.stattests import ks_stat_test, chi_stat_test, z_stat_test
+from evidently.analyzers.stattests.registry import get_stattest, stattest, StatTestInvalidFeatureTypeError, \
+    StatTestNotFoundError
+
+
+@stattest("test_stattest", allowed_feature_types=["cat"])
```
Tapot, line 8: This code changes global stat tests, maybe better will be move it to a test method, save initial state or remove test functions after the test run?

src/evidently/analyzers/stattests/registry.py
Tapot, line 8: Do we want to raise an error for collisions in test names?

PR #199, opened 2022-04-12
add python command to grafana integration example

nathnx:
add python command to execute dataset creation in the grafana integration example


emeli-dral:
Now it looks better, thank you🙏

PR #204, opened 2022-04-20
revise kdd dataset name for the Grafana integration example

nathnx:
- change the default value of the global variable `KDD_DATASET_NAME` to match with the data source name
- revise some code formatting


Tapot (approved):

Tapot:
> * change the default value of the global variable `KDD_DATASET_NAME` to match with the data source name
> * revise some code formatting

Thank you for the update!

PR #220, opened 2022-05-04
Replace feature_stattest_func with new options.

Liraim:
Replace feature_stattest_func with stattest_func, cat_stattest_func, num_stattest_func, per_feature_stattest_func.
Old parameter left intact with warning on usage and exception if used along with new options.

### DIFF
--- a/src/evidently/options/data_drift.py
+++ b/src/evidently/options/data_drift.py
@@ -11,6 +11,33 @@ DEFAULT_NBINSX = 10
 
 @dataclass
 class DataDriftOptions:
+    """ Configuration for Data Drift analyzer, Dashboard and Profile.
+
+    Attributes:
+        confidence: Defines the confidence level for statistical tests.
+                    Applies to all features (if passed as float) or certain features (if passed as dictionary).
+                    (Deprecated) Use `threshold` to define confidence level for statistical
+                     tests as more universal solution.
+        threshold: Defines thresholds for statistical tests.
+                   Applies to all features (if passed as float) or certain features (if passed as dictionary).
+        drift_share: Sets the share of drifting features as a condition for Dataset Drift in the Data Drift report.
+        nbinsx: Defines the number of bins in a histogram.
+                Applies to all features (if passed as int) or certain features (if passed as dictionary).
+        xbins: Defines the boundaries for the size of a specific bin in a histogram.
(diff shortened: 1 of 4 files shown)


src/evidently/options/data_drift.py
```diff
@@ -16,7 +16,14 @@ class DataDriftOptions:
     drift_share: float = 0.5
     nbinsx: Union[int, Dict[str, int]] = DEFAULT_NBINSX
     xbins: Optional[Dict[str, int]] = None
+
     feature_stattest_func: Optional[Union[PossibleStatTestType, Dict[str, PossibleStatTestType]]] = None
+
+    stattest_func: Optional[PossibleStatTestType] = None
```
Tapot, line 22: can you add some descriptions in docstrings or comments about the parameters?

src/evidently/options/data_drift.py
```diff
@@ -16,7 +16,14 @@ class DataDriftOptions:
...
+    cat_stattest_func: Optional[PossibleStatTestType] = None
+    num_stattest_func: Optional[PossibleStatTestType] = None
+    per_feature_stattest_func: Optional[Dict[str, PossibleStatTestType]] = None
```
Tapot, line 25: Maybe we can use `feature_stattest_func` and do not add a new parameter?

src/evidently/options/data_drift.py
Liraim: done

src/evidently/options/data_drift.py
Liraim: Old options left there for backward compatibility as we decided to change it behaviour and rename all new options for simplicity.

tests/options/test_data_drift.py
```diff
@@ -54,6 +54,15 @@ def _another_stattest():
     pass
 
 
+_features = dict([("cat1", "cat"), ("cat2", "cat"), ("num1", "num"), ("num2", "num")])
+
+
+def _features_st(*feature_func):
```
Tapot, line 60: I think it is too complicate code for tests - it is hard to read test cases with it

Can you rewrite it with explicit dicts in parametrize or at least add docs and comments for the method?

Tapot (approved):

tests/options/test_data_drift.py
Liraim: Simplified test a bit.

PR #221, opened 2022-05-06
Update setup.py

Riya-Anna-Jeny:


Liraim:
Hi @Riya-Anna-Jeny ,
Thanks for your interest in our library.
Can you provide short explanation about this PR? Why do you think we need change version of plotly dependency?

emeli-dral:
Hey @Riya-Anna-Jeny ,
Thank you for your path!
For now we decided to keep plotly version in setup.py. 
If you believe that it was wrong decision,  feel free to reopen this pull request and comment on the topic 🙏

PR #233, opened 2022-05-19
fixed DataQualityMonitorMetrics feature_name return value

obbiondo:
Fix DataQualityMonitorMetrics yielding wrong value for feature name.

It currently returns feature_type instead of feature_name.


Liraim (approved):
Thanks for your contribution.

PR #244, opened 2022-06-02
added dash to docker compose to README.md and run_example.py script

jeff-ridgeway:
Submitting a pull request to add a dash to the README.md under the grafana integration examples as well as the `run_example.py`. There is a dash that is needed between `docker` and `compose` for `docker-compose`. Not having the dash for docker compose makes it so that the `run_example.py` errors out before building out the grafana and Prometheus dashboard.

### DIFF
--- a/examples/integrations/grafana_monitoring_service/README.md
+++ b/examples/integrations/grafana_monitoring_service/README.md
@@ -81,7 +81,7 @@ To stop the process of sending data, cancel the execution of the example script 
 If the Docker containers were not stopped after that, run a command:
 
 ```bash
-docker compose down
+docker-compose down
 ```
 
 ## How to customize the Grafana Dashboard view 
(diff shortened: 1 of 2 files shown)


Tapot:
Hello, in Docker docs two ways for running Docker compose are mentioned - via a separate `docker-compose` as an old-way and a new command `docker compose` https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command

We choose to use a new way, a separate command in `docker`

May be you can describe reasons for the replacement it to `docker-compose`?

emeli-dral:
Hi @jeff-ridgeway ,
Thank you for bringing this up.
We decided to keep using compose V2 , which is described here: https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command 
If you believe it is wrong decision and we need to reconsider it, please feel free to comment on that and reopen this pull request.

PR #247, opened 2022-06-09
Small changes/fixes for `CONTRIBUTING.md` instructions

strickvl:
- make capitalisation of 'Pull Request' uniform
- fix old / non-working instructions about `flake8`
- fix one small typo

I did also notice that the instruction to run `pip install -e .[dev]` did not work for me. I got an error that `dev` was not found, I think. So I just removed the `[dev]`, then installed the rest of the dev dependencies manually. There's probably something to be fixed in this document about that, but I'm not sure what it is. I welcome suggestions from your side on that point.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 Thank you for considering contributing to Evidently!
 
 ## How can you contribute?
-We welcome both code and none-code contributions. You can:
+We welcome both code and non-code contributions. You can:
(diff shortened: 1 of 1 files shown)


Liraim:
About `pip install -e .[dev]` in some shells you need to escape `.[dev]` part, for example on MacOS in zsh you would need to type `pip install -e '.[dev]'`

Liraim (approved):

PR #249, opened 2022-06-11
Fix docker execution in grafana_monitoring_service

Utku-Savas:
I changed docker compose up -d command to docker-compose up in run_examples.py file and updated README


Liraim:
Hi,
`docker-compose` (aka Docker Compose v1) is deprecated and now you should use `docker compose` (aka Docker Compose v2). You should check instructions about https://docs.docker.com/compose/install/ .

PR #252, opened 2022-06-14
Update README.md

amarutha:
In the Stop the Examples Section
Please include this 
docker-compose -f ./examples/integrations/grafana_monitoring_service/docker-compose.yml down


Tapot:
Hello, in Docker docs running `docker compose` is described as a main way for Docker Compose launch. AN if you do in from the example folder, it can find the config file.

Maybe you can explain why do you want to update the readme file this way?

PR #255, opened 2022-06-22
Predefined color schemas for `Dashboard`

strickvl:
I added four predefined color schemas which users can use out of the box.

I also updated the docs to show how to use them.

(I thought it wise not to include too many predefined options (as then you have choice paralysis), and I thought maybe it could be a good idea to give them quirky / memorable names.)

Feedback welcomed :)


strickvl:
Hmmm failing tests! The `mypy` docs seem to suggest that changing `Tuple[str]` to `tuple[str]` should fix the problem, but it doesn't seem to fix it when I try it locally. I'll take another look at this later.

PR #274, opened 2022-07-13
Update utils.py

polaser:
Use unique() method instead of function set() over object of Series
On big objects of Series set() is TOO SLOW


polaser:
@0lgaF I saw that you are quite involved in developing this tool

sorry for disturbing 

we are trying to use evidently on our production stage but it is necessary to boost some calculations 

could you please review these changes?

Liraim:
Hi, Thanks for your contribution.

PR #290, opened 2022-08-03
install pre release version

kemingy:
It seems this package only has pre-release versions in [PyPI](https://pypi.org/project/evidently/#history).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ There is a lightweight integration with Prometheus and Grafana that comes with p
 ### MAC OS and Linux
 Evidently is available as a PyPI package. To install it using pip package manager, run:
 ```sh
-$ pip install evidently
+$ pip install --pre evidently
 ```
 If you want to generate reports as HTML files or export as JSON profiles, the installation is now complete.
 


Tapot:
> It seems this package only has pre-release versions in [PyPI](https://pypi.org/project/evidently/#history).

Hello!

THank you for the suggestion!

But now we have only pre-release versions for the lib, an it is ok to install the latests version with just
`pip install`, without `--pre` flag. You can take a look in [this pip tool](https://github.com/pypa/pip/issues/5503) issue and [PEP 440 section](https://peps.python.org/pep-0440/#handling-of-pre-releases) for more details.

Later, when we will make a stable release, the docs will be the same and not out-of-date - the default way to install should not include `--pre` flag. And than we will add more docs for pre-release and dev versions installation.

If you have another reason for adding the flag or an issue that you face without it - it will be interesting to read in a comment or description.

But now we want to leave the installation instruction without `--pre` flag in the docs.

kemingy:
Hi @Tapot thanks for your explanation.

Previously I'm using another Python package manager tool and it failed to get the latest version if I don't pin the Python minimum required version. So I think it will be more clear if `--pre` is passed explicitly.

I see your point. Closing this one.

Tapot:
> So I think it will be more clear if --pre is passed explicitly.
Yeah, we hope that it will became more clear after the stable version will be released)

Issue #347, opened 2022-09-29
Hellinger distance for Drift Analysis


emeli-dral:
Hellinger distance: https://en.wikipedia.org/wiki/Hellinger_distance
How to contribute stattest for Drift Analysis: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest

- [x] test implementation 
- [x] unittest
- [x] docs update 
- [x] example update

Mathanraj-Sharma:
Hi, @emeli-dral I would like to take this, 

But I am new to evidently, could you please assign this to me and help me during the progress?

Are there any time constraints you have, do I need to complete it within a short time from assigning it to me?

inderpreetsingh01:
Hi @Mathanraj-Sharma I already started working on this issue yesterday, forgot to mention it here, @emeli-dral please review this pr, will add tests and example shortly.

emeli-dral:
Hi @Mathanraj-Sharma!

Thanks for your interest to contribute! We do not have a specific timeline here and will not formally "reserve" issues - though we encourage everyone to post when they start working on the issue so that others can see it (and maybe pick a different test). We will review contributions as they come, on the first come - first reviewed basis.

Note that you can also choose and implement a different statistical test of your choice!

emeli-dral:
We recently added an [example](https://github.com/evidentlyai/evidently/blob/main/examples/how_to_questions/how_to_specify_stattest_for_a_testsuite.ipynb) of setting stattests in TestSuites. We are happy to receive PRs to update it with new drift detection methods!

PR #352, opened 2022-10-02
added hellinger_distance for drift detection

inderpreetsingh01:
Adds **Hellinger Distance** drift detection metric as mentioned in #347 

Also fixes one typo in get_binned_data in src/evidently/calculations/stattests/utils.py


emeli-dral:
Hi @inderpreetsingh01,
Thank you for your job! 

The statistical test looks good, great decision to make it normed.

Here is one thing we suggest to change: could you please delete files index.js and index.js.LICENSE.txt? These files are usually generated automatically when the package is created and we do not track them in the repos.

And I haven’t seen any unittests yet. Will you add them to this PR later?
Or do you prefer to send them separately?

Issue #343, opened 2022-09-29
Cramér-von Mises (CVM) stattest for Drift Analysis


emeli-dral:
Cramér-von Mises Criterion: https://en.wikipedia.org/wiki/Cram%C3%A9r%E2%80%93von_Mises_criterion
How to contribute stattest for Drift Analysis: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest

- [x] test implementation 
- [x] unittest
- [x] docs update 
- [x] example update

emeli-dral:
Oh, I think I need to mention that there is already a PR on this specific test by @SangamSwadik waiting for minor fixes. 
But you can also choose a different test to implement, including a custom one if you use something different in your practice!

ReubenThomasJohn:
Ah okay! Will see if there are any other stattests that need to be implemented. Thank you! @emeli-dral

emeli-dral:
We recently added an [example](https://github.com/evidentlyai/evidently/blob/main/examples/how_to_questions/how_to_specify_stattest_for_a_testsuite.ipynb) of setting stattests in TestSuites. We are happy to receive PRs to update it with new drift detection methods!

sankalp250:
I’d like to work on implementing the Cramér‑von Mises drift test. Could I be assigned?

PR #356, opened 2022-10-04
Add Cramér-von-mises test

SangamSwadiK:
**Issue:** 
#343 (Add Cramer von mises stattest)

**What does this implement/fix?**
Adds Cramer von mises and its test

@emeli-dral 
Please let me know if any improvements or changes are needed.
Thanks


emeli-dral:
Hey @SangamSwadiK ,
thanks for one more test!

The thing is, that we still want to support Python 3.6, but starting from the 1.6.0 Scipy requires at least Python 3.7
https://scipy.github.io/devdocs/release.1.6.0.html 
That is the reason, why we keep 1.54 version of Scipy as a dependency.

Issue #344, opened 2022-09-29
Anderson–Darling stat test for Drift Analysis


emeli-dral:
Anderson–Darling test: https://en.wikipedia.org/wiki/Anderson%E2%80%93Darling_test 
How to contribute stattest for Drift Analysis: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest

- [x] test implementation 
- [x] unittest
- [x] docs update 
- [x] example update

AntonisCSt:
Hi, Is it possible to be assigned for this enhancement? I am new to contributing to open source (I also joined hackoctoberfest)

emeli-dral:
We recently added an [example](https://github.com/evidentlyai/evidently/blob/main/examples/how_to_questions/how_to_specify_stattest_for_a_testsuite.ipynb) of setting stattests in TestSuites. We are happy to receive PRs to update it with new drift detection methods!

PR #358, opened 2022-10-04
Add anderson in statistical test options

SangamSwadiK:
**Issue**: #344 

**changes implemented ?**
Add `anderson` in the  test options.


emeli-dral (changes requested):
Thanks for updating docs, we appreciate it!

The only thing I suggest to update here is the test name in the description: from "Anderson darling" to "Anderson–Darling".

emeli-dral (approved):

PR #363, opened 2022-10-06
added Epps-Singleton test

AntonisCSt:
This is a contribution from Hacktoberfest 2022.

I added the Epps-Singleton test.

+added test function
+added test for pytest
+changed documents

The method was borrowed by scipy library:  [ES link scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.epps_singleton_2samp.html#scipy.stats.epps_singleton_2samp)

It was not suggested originally by the Issues, so I added it myself.

Why use it?

_**"One advantage of the ES test compared to the KS test is that is does not assume a continuous distribution. In [[1]](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.epps_singleton_2samp.html#r8ab16cf236a2-1), the authors conclude that the test also has a higher power than the KS test in many examples. They recommend the use of the ES test for discrete samples as well as continuous samples with at least 25 observations each, whereas [anderson_ksamp](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.anderson_ksamp.html#scipy.stats.anderson_ksamp) is recommended for smaller sample sizes in the continuous case."**_


AntonisCSt:
Hi in this case of conflict (after review was requested), should I edit locally and merge again? Thanks

emeli-dral:
Hi @AntonisCSt ,
Yes, if you merge main in your branch and resolve conflicts it will be the fastest way to proceed.

emeli-dral:
Hi @AntonisCSt 

I tried to run Epps-Singleton test with almost similar reference and current data (that’s what often happens, when users analyse very similar datasets), and I got an error with SVD.

Here are the data I had:
```
a = [1] * 30
b = [1]*25 + [1.3,1.2,1.1,1.05,1.02]
```

And that is how I ran the example:
```
_epps_singleton(a, b, feature_type='num', threshold=0.1)
```

I got the following error:
```
LinAlgError                               Traceback (most recent call last)
Input In [23], in <cell line: 1>()
----> 1 _epps_singleton(a, b, feature_type='num', threshold=0.1)

Input In [4], in _epps_singleton(reference_data, current_data, feature_type, threshold)
      1 def _epps_singleton(
      2     reference_data: pd.Series,
      3     current_data: pd.Series,
      4     feature_type: str,
      5     threshold: float,
      6 ) -> Tuple[float, bool]:
      7     """Run the Epps-Singleton (ES) test of two samples.
      8     Args:
      9         reference_data: reference data
   (...)
     14         test_result: whether the drift is detected
     15     """
---> 16     p_value = epps_singleton_2samp(reference_data, current_data)[1]
     17     return p_value, p_value < threshold

File /opt/homebrew/Caskroom/miniconda/base/envs/evidently53/lib/python3.9/site-packages/scipy/stats/_hypotests.py:128, in epps_singleton_2samp(x, y, t)
    126 cov_y = np.cov(gy.T, bias=True)
    127 est_cov = (n/nx)*cov_x + (n/ny)*cov_y
--> 128 est_cov_inv = np.linalg.pinv(est_cov)
    129 r = np.linalg.matrix_rank(est_cov_inv)
    130 if r < 2*len(t):

File <__array_function__ internals>:180, in pinv(*args, **kwargs)

File /opt/homebrew/Caskroom/miniconda/base/envs/evidently53/lib/python3.9/site-packages/numpy/linalg/linalg.py:1998, in pinv(a, rcond, hermitian)
   1996     return wrap(res)
   1997 a = a.conjugate()
-> 1998 u, s, vt = svd(a, full_matrices=False, hermitian=hermitian)
   2000 # discard small singular values
   2001 cutoff = rcond[..., newaxis] * amax(s, axis=-1, keepdims=True)

File <__array_function__ internals>:180, in svd(*args, **kwargs)

File /opt/homebrew/Caskroom/miniconda/base/envs/evidently53/lib/python3.9/site-packages/numpy/linalg/linalg.py:1657, in svd(a, full_matrices, compute_uv, hermitian)
   1654         gufunc = _umath_linalg.svd_n_s
   1656 signature = 'D->DdD' if isComplexType(t) else 'd->ddd'
-> 1657 u, s, vh = gufunc(a, signature=signature, extobj=extobj)
   1658 u = u.astype(result_t, copy=False)
   1659 s = s.astype(_realType(result_t), copy=False)

File /opt/homebrew/Caskroom/miniconda/base/envs/evidently53/lib/python3.9/site-packages/numpy/linalg/linalg.py:98, in _raise_linalgerror_svd_nonconvergence(err, flag)
     97 def _raise_linalgerror_svd_nonconvergence(err, flag):
---> 98     raise LinAlgError("SVD did not converge")

LinAlgError: SVD did not converge
```

Could you please take a look?

AntonisCSt:
Thanks for finding this issue @emeli-dral !!!

I ran your data with the  Epps-Singleton and had the same issue. 🤓

This issue comes from scipy's Epps-Singleton function.

Specifically: It comes from the calculation of [Interquartile range (IQR)](https://github.com/scipy/scipy/blob/656076ca6b490f587e9bd9c4cd10cb259a687c5b/scipy/stats/_hypotests.py#L117).
For this particular data, the iqr is zero which then makes the [ts](https://github.com/scipy/scipy/blob/656076ca6b490f587e9bd9c4cd10cb259a687c5b/scipy/stats/_hypotests.py#L118) infinite.

So it looks like that will happen mostly when we have the same value repeated ([1] * 30).
(You can also get this error with nan values I think)

So, how do you suggest proceeding?:

1) Report the issue to scipy and find a solution?
2) Create my own epss-singleton?
3) Merge with a raised error?
4) any other idea?

emeli-dral:
Hey @AntonisCSt ,
yes, probably it is better to just choose another stat test for such kind of data.
Moreover, data in my example is something in between numerical and categorical 🙂 Especially ```a = [1] * 30```

I believe the best next step would be to merge with a raised error. 

Could you please merge main into your branch as well?

PR #366, opened 2022-10-07
Add G-stattest and its test

SangamSwadiK:
**What does this PR implement ?**
Proposes to add a g-test for categorical features.
- [x] test implementation
- [x] unittest
- [ ] doc
- [ ]  examples

**When/Why use this ?**
G ≈ χ2 when the observed counts  are close to the expected counts When this difference is large, however, the χ2 approximation of begins to break down. Here, the effects of outliers in data will be more pronounced, and this explains the why  tests fail in situations with little data.

**Refs:-**
[https://en.wikipedia.org/wiki/G-test](https://en.wikipedia.org/wiki/G-test)
[https://stats.stackexchange.com/questions/289261/whats-the-usage-difference-between-g-test-and-chi-square-test](https://stats.stackexchange.com/questions/289261/whats-the-usage-difference-between-g-test-and-chi-square-test)

### DIFF
--- a/tests/stattests/test_stattests.py
+++ b/tests/stattests/test_stattests.py
@@ -6,8 +6,8 @@ from pytest import approx
 from evidently.calculations.stattests import z_stat_test
 from evidently.calculations.stattests.anderson_darling_stattest import anderson_darling_test
 from evidently.calculations.stattests.chisquare_stattest import chi_stat_test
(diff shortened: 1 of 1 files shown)


emeli-dral:
Hey @SangamSwadiK, 
Thank you!

Issue #334, opened 2022-09-26
The fixed value for feel_zeroes in get_binned_data may lead to deviation in some case.


chi2liu:
In the implemntation of get_binned_data https://github.com/evidentlyai/evidently/blob/4e4637d361d11354d26f361a07e1c5a847bbfc74/src/evidently/calculations/stattests/utils.py#L33

By default, 0.0001 is used to fill the value with a percent of 0. It is valid when the minimum percent of non-zero is greater than 0.0001. But when the minimum percent in the data is less than or equal to 0.0001, like 0.00001. After we filling zeros, the original zero percent is now 0.0001, which is larger than the existing minimum percent(0.00001), and the data distribution is wrong after filling zeros.

The default value of fill zeros in this place, I think, should be set dynamically according to the percentage of data, rather than being fixed directly. How to set dynamically needs further discussion. For example, it should be set to one tenth of the minimum percent, or other values, but it must be guaranteed to be smaller than the minimum percent, otherwise it will be wrong.

I found this problem because when I tried to calculate Kullback Leibler divergence drift score on my data, I found that the minimum percent of my data was 0.00001, which was smaller than the default fill zeros, which resulted in incorrect Kullback Leibler divergence drift score. For example, the default calculated Kullback Leibler divergence drift score in my data is 0.41. After I change fill zeros to one tenth of the minimum percent, the calculated Kullback Leibler divergence drive score is 1.9.

emeli-dral:
Hi @chi2liu,

Thanks for reporting it! We will not be able to address it immediately as the ongoing API update is currently a priority, but I added it to the list of feature requests we regularly review. Will get there!  

We also welcome a contribution if you want to fix it!

AntonisCSt:
Wow, I managed to mess up a bit with this issue by adding many different references in my local commits! 
Sorry!!
(new to contributing here)

I added opened a pull request for this with a temporal change.

I added an if statement:

```
np.place(reference_percents, reference_percents == 0, min(reference_percents[reference_percents!=0])/10**6 if min(reference_percents[reference_percents!=0]) <= 0.0001 else 0.0001)
np.place(current_percents, current_percents == 0, min(current_percents[current_percents!=0])/10**6 if min(current_percents[current_percents!=0]) <= 0.0001 else 0.0001)
```

So, I basically did what you mentioned @chi2liu but instead of one-tenth I added 10^6. This is just to make it as close to zero (the actual value).

[...]

PR #371, opened 2022-10-09
fix for issue #334 feel zeroes of stattests utils

AntonisCSt:
#334 

Temporarily added a fix for the issue mentioned by @chi2liu.

Now if the minimum percent in the data is less than or equal to 0.0001, the zeroes will be converted to minimum /10^6.

10^6 was chosen to have the same effect as 0.0001 has when the minimum is not less than or equal to 0.0001.

This needs more discussion because if you increase the 10^6, the drift metrics will also increase. 
In my opinion, they should be as close to zero as we can (compared to the other values).

check this notebook for an example that you can run: https://github.com/AntonisCSt/evidently/blob/example-utils/src/utils_feel_zeroes_issue_fix.ipynb

### DIFF
--- a/src/evidently/calculations/stattests/utils.py
+++ b/src/evidently/calculations/stattests/utils.py
@@ -35,7 +35,7 @@ def get_binned_data(
         current_percents = np.array([current_feature_dict[key] / len(current_data) for key in keys])
 
     if feel_zeroes:
-        np.place(reference_percents, reference_percents == 0, 0.0001)
-        np.place(current_percents, current_percents == 0, 0.0001)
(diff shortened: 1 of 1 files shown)


AntonisCSt:
Excuse me for the linter/black fail tests, but If I add `black [file-path] ` it changes the whole file. (based on the [Contribution guide](https://github.com/evidentlyai/evidently/blob/main/CONTRIBUTING.md) )

should this happen? or do you have a  for example .toml file specifying the black settings?

Tapot:
> should this happen? or do you have a for example .toml file specifying the black settings?

black and isort setting are specified in setup.cfg, launch `black .` and `isort .` from the repo's root should help

emeli-dral:
Hi @AntonisCSt ,

we agree with the idea to chose 10**6 to have the same effect as 0.0001 has when the minimum is not less than or equal to 0.0001. 

It worth future exploring indeed, although that's already good fix. We merge it.

Issue #345, opened 2022-09-29
Fisher’s Exact test for Drift Analysis


emeli-dral:
Fisher’s Exact test: https://en.wikipedia.org/wiki/Fisher%27s_exact_test 
How to contribute stat test for Drift Analysis: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest

- [ ] test implementation 
- [ ] unittest
- [ ] docs update 
- [ ] example update

mertbozkir:
I would like to assign this issue to myself. I want to contribute evidently! 🤗

emeli-dral:
Hi @mertbozkir! Thanks for your interest! 
Feel free to start working on the issue, we will review pull requests as they come!

emeli-dral:
We recently added an [example](https://github.com/evidentlyai/evidently/blob/main/examples/how_to_questions/how_to_specify_stattest_for_a_testsuite.ipynb) of setting stattests in TestSuites. We are happy to receive PRs to update it with new drift detection methods!

mertbozkir:
@SangamSwadiK asked below question in the Discord server 

> Hi, for the  fisher's exact test in the issues, would fisher for 2x2 contingency table suffice or should it be able to work for m x n contingency table ?

PR #373, opened 2022-10-10
Add  fisher's exact

SangamSwadiK:
**Issue:** 
#345 (Add fisher's exact test for binary data)

**What does this implement/fix ?**
This adds fisher's exact test, its doc and necessary changes in the How to examples.

- [x] test implementation
- [x] unittests
- [x] doc
- [x]  examples

cc @mertbozkir


emeli-dral:
Hey @SangamSwadiK, 

The test looks very good! 

It is designed to work with binary data, and I have a corner case, which I believe worth supporting.
Here is the example. Imagine I have a binary feature and suddenly one of two values disappeared from the data? In this case it is still binary data, but test fails on such an input:

```
_fisher_test([1,0,1,0,1], [0,0,0,0,0], feature_type='cat', threshold=0.1)

ValueError                                Traceback (most recent call last)
Input In [40], in <cell line: 1>()
----> 1 _fisher_test([1,0,1,0,1], [0,0,0,0,0], feature_type='cat', threshold=0.1)

Input In [34], in _fisher_test(reference_data, current_data, feature_type, threshold)
      4 """Calculate the p-value of Fisher's exact test between two arrays
      5 Args:
      6     reference_data: reference data
   (...)
     12     test_result: whether the drift is detected
     13 """
     15 contingency_matrix = pd.crosstab(reference_data, current_data)
---> 17 _, p_value = fisher_exact(contingency_matrix)
     19 return p_value, p_value < threshold

File /opt/homebrew/Caskroom/miniconda/base/envs/evidently53/lib/python3.9/site-packages/scipy/stats/_stats_py.py:4278, in fisher_exact(table, alternative)
   4276 c = np.asarray(table, dtype=np.int64)
   4277 if not c.shape == (2, 2):
-> 4278     raise ValueError("The input `table` must be of shape (2, 2).")
   4280 if np.any(c < 0):
   4281     raise ValueError("All values in `table` must be nonnegative.")

ValueError: The input `table` must be of shape (2, 2).

```

Could you please take a look and support this scenario? 

I also suggest to raise a ValueError that will appear if you have more than 2 unique values in the input data and explicitly tell the user that the test should be applied to the binary data only. I think it might be more straightforward comparing to ```ValueError: The input `table` must be of shape (2, 2).``` What do you think?

emeli-dral:
Hi @SangamSwadiK,

Thank you so much for changes you made! 
I like a lot how you implemented exact test, and especially your approach for testing not only stattest function but utils as well. 

I have a question about corner cases we discussed before. 
Here is my example:
```
_fisher_exact_stattest(pd.Series(["a","a","a","a","a","a","a","a","a","a"]), 
                       pd.Series(["b","b","b","b","b","b","b","b","b","b"]), feature_type='cat', threshold=0.1)
(1.0, False)
```

I see that there are no fails, which is great! But I'm a little surprised that we do not detect a drift here.
What do you think?

emeli-dral:
Hi @SangamSwadiK, 

I have a suggestion how to act on it.

Indeed the test itself will not compute in case we have only 1 category in reference, and 1 (different) category in current. 

However this might indeed happen in practice when you deal with binary features. From the user standpoint, when this happens, it is something you want to know about. If all values are different between reference and current, we can safely assume that the feature has drifted. 

I would suggest to implement this case separately. Instead of returning the value error, we can return the “drift detected” outcome when all values are different (without actually computing the statistical test).

emeli-dral:
I am also labelling the issues as HF-accepted to acknowledge all the work done!

emeli-dral:
Hi @SangamSwadiK ,

I checked the latest version and I like it a lot! 

"I didn't write them explicitly because the p-values vary a lot with different lengths of both reference and current even though they are flipped entirely" - I agree with your approach. I think it is better to return p-value here, it might be useful for something like drift size comparison, when users need to order features by drift size and act on top of this order.

Thank you a lot for your contribution! I'm happy to merge 🥳

___
@elenasamuylova this is a very nice example of how pytests should be implemented.

PR #380, opened 2022-10-12
Mann-Whitney U-rank drift test

AntonisCSt:
This is a contribution from Hacktoberfest 2022.

I added the Mann-Whitney U-rank drift test.

+added test function
+added test for pytest
+changed documents

The method was borrowed by scipy library: [MW link scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mannwhitneyu.html#r31b0b1c0fec3-4)

It was not suggested originally by the Issues, so I added it myself.

Why use it?

"Unlike the independent-samples t-test, the Mann-Whitney U test allows you to draw different conclusions about your data depending on the assumptions you make about your data's distribution. These conclusions can range from simply stating whether the two populations differ through to determining if there are differences in medians between groups. These different conclusions hinge on the shape of the distributions of your data."

That means also that If you have extremely different sizes of data this will also count as a drift :)

Sources: 
[statistics.laerd](https://statistics.laerd.com/spss-tutorials/mann-whitney-u-test-using-spss-statistics.php)
[MIT lecture 5.1.4 Mann-Whitney U Test ](https://www.mit.edu/~6.s085/notes/lecture5.pdf)

I also have an example notebook that I can add :)


emeli-dral:
Hi @AntonisCSt ,

I went through the implementation, test works well!

I request a little change in the naming. Could you please rename ```mannwhitneyu_rank``` to ```_mannwhitneyu_rank```?
This underscore at the beginning shows that the function is meant to be used  internally, which is our case exactly .

Issue #378, opened 2022-10-11
Add a new data drift detection method to Evidently


emeli-dral:
This is an open issue. We invite you add new data drift detection tests or metrics to the Evidently library. 

Here are general instruction on how to add a new test:
https://github.com/evidentlyai/evidently/wiki/Hacktoberfest 

We invite you contribute new drift detection methods that you consider useful. If you want to add a new method, you can create a separate issue to let us know that you are working on it. You can also directly send the pull-request. 

Please do share some details about the new method and when it can be useful (or when you are using it!) 

Each contribution should contain:
- Test implementation
- Unittest
- Docs update
- Example update

emeli-dral:
Our goal is not to implement as many statistical tests as possible (as the choice can be confusing to the users!) but to add tests that have different properties and might be chosen over already existing ones for a specific reason. 

We highly welcome additions that come from your practice: e.g., if you, as a data scientist, already use a different drift detection method in your work. 

Here are some ideas for methods you can explore:
- Bhattacharyya distance https://en.wikipedia.org/wiki/Bhattacharyya_distance 
- Energy distance ​https://en.wikipedia.org/wiki/Energy_distance ​ 
- Total variation https://en.wikipedia.org/wiki/Total_variation 
- Matrix or vector norm https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html 
 
You can also study implementations that exist in research papers and propose ones that could be a valuable addition. Please cite the paper if you implement any of the approaches proposed by the authors so that we reflect it in the documentation.

aswinvk28:
Mahalanobis distance is also another candidate. 

Here is a paper that discusses about Mahalanobis distance and its applications in Outlier detection
[https://core.ac.uk/download/pdf/233075917.pdf](https://core.ac.uk/download/pdf/233075917.pdf)

PR #400, opened 2022-10-19
Energy distance num drift test

AntonisCSt:
#378 

Added Energy distance proposed by issue 378:

This one was very interesting! It was influenced by **_Newton's [gravitational potential](https://www.sciencedirect.com/topics/mathematics/gravitational-potential) energy, and there is an elegant relation to the notion of potential energy between statistical observations._**

While there were a lot of theoretical sources, apart from scipy's library (from which I borrowed the methodology), I could not find many practical ones.

I tested on multiple distributions (drifted cases and non-drifted) and found that the threshold of 0.1 is satisfying.

If you have any other useful resources please let me know :)

Enjoy for now!

sources: [wikipedia Energy distance](https://en.wikipedia.org/wiki/Energy_distance)
[Scipy library](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.energy_distance.html#id1)
[Energy statistics: A class of statistics based on distances](https://doi.org/10.1016/j.jspi.2013.03.018)

### DIFF
--- a/docs/book/customization/options-for-statistical-tests.md
+++ b/docs/book/customization/options-for-statistical-tests.md
@@ -180,3 +180,7 @@ example_stat_test = StatTest(
   - only for numerical features
   - returns `p_value`
   - drift detected when `p_value < threshold`
+- `ed` - Energy distance
+  - only for numerical features
+  - returns `distance`
+  - drift detected when `distance >= threshold`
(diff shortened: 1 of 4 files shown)


emeli-dral:
Hi @AntonisCSt ,

nice distance, I enjoyed reading 🙂
It will be great to have it, thank you for the contribution!

Could you please solve conflicts, so that we can merge the distance?

PR #401, opened 2022-10-19
Added drift stattest comparison notebook 

AntonisCSt:
Hi,

**What?**
After working a bit with some tests, and the suggestion by @emeli-dral from the [Evidently Hacktoberfest Community Call](https://www.youtube.com/watch?v=DyzKNXNIfGA) I added a short notebook that compares tests based on the strong and weak points.

It focuses on the recent  Mann-Whitney U-rank test and the Kolmogorov–Smirnov tests. It gives a couple of examples of when one is better than the other. 

**Why?**
The overall conclusion wanted to be: "Choose your test wisely" (based on their strong points :D)

It can definitely be enriched with more examples. Hope it will be useful.

Please let me know if it can be merged or if you would like something else :)


AntonisCSt:
The test fails because of Seaborn. Any suggestion on how to use seaborn in the notebook? Or should I do it in a different way?

emeli-dral:
@AntonisCSt Indeed it looks like tests fail, because there are no Seaborn installed.
Could you please try to add Seaborn to  ```dev_requirements.txt``` ?

AntonisCSt:
> @AntonisCSt Indeed it looks like tests fail, because there are no Seaborn installed. Could you please try to add Seaborn to `dev_requirements.txt` ?

@emeli-dral Thanks!! I just added it :)

emeli-dral:
@AntonisCSt Thank you!

PR #484, opened 2022-12-17
Update requirements to exclude dataclasses from python3.7+

AugustKarlstedt:
from https://pypi.org/project/dataclasses/
> As of version 0.4, this code no longer works with Python 3.7. For 3.7, use the built-in dataclasses module.


emeli-dral:
Hi @AugustKarlstedt ,
thank you!

PR #498, opened 2023-01-30
Fix typo in Docs

SangamSwadiK:
Fix typo in z_stat_test example usage.
Also, the API reference needs to be fixed as well.
https://docs.evidentlyai.com/reference/api-reference/evidently.calculations/evidently.calculations.stattests#module-evidently.calculations.stattests.z_stattest


emeli-dral:
Hey @SangamSwadiK ,
Thanks for the fix!

Issue #494, opened 2023-01-17
Jensen-Shannon distance should be computed using base 2 logarithms for it to be bound between 0 and 1


perezbecker:
Your [current implementation for the Jensen-Shannon Distance (JSD)](https://github.com/evidentlyai/evidently/blob/main/src/evidently/calculations/stattests/jensenshannon.py) is leveraging [scipy.spatial.distance.jensenshannon](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.jensenshannon.html), which defaults to computing the JSD using logarithms with base e. Under this definition, the JSD is bound between 0 and sqrt(ln(2))=0.83255... (see, for example, the [JSD Wikipedia article](https://en.wikipedia.org/wiki/Jensen%E2%80%93Shannon_divergence))

I believe you want to compute JSD using base 2 logarithms, so that it is bound between 0 and 1 as you state in your [data drift detection blog post](https://www.evidentlyai.com/blog/data-drift-detection-large-datasets).

Here is code which reproduces this issue by computing the JSD between two extremely drifted distributions: 
```
from evidently.calculations.stattests import jensenshannon_stat_test
import numpy as np
import pandas as pd

x = pd.Series(np.random.normal(100, 10, 100_000))
y = pd.Series(np.random.normal(10_000, 10, 100_000))

print(jensenshannon_stat_test(x, y, feature_type='num',threshold=0.1))
>>> StatTestResult(drift_score=0.8325546111576977, drifted=True, actual_threshold=0.1)
```
Note that the `drift_score` is close to the current maximum theoretical value of `sqrt(ln(2))=0.83255...`, instead of the desired value of `1`.

emeli-dral:
Hi @perezbecker ,

Thank you for mentioning that! We have so many nonnormalized drift detection methods (like PSI) that we have given up trying to normalize everything🙈 But in vain! :) I agree that many users will be happy to use the normalized JS test. The only thing is that we will keep the base e by default for now because the selected default thresholds for drift detection do not take into account normalization.

emeli-dral:
Hey @SangamSwadiK ,
Thank you for the contribution! We will merge it soon:)

emeli-dral:
Hi @perezbecker,

We recently merged the update to JensenShannon distance. Now ```_jensenshannon``` accept logarithm base as a parameter (Thanks to @SangamSwadiK !).

[...]

PR #499, opened 2023-01-30
add base parameter in JensonShannon distance

SangamSwadiK:
**Issue ::**   #494 

**What does this fix ?** :  adds a base parameter (defaults to base2)

Thanks!


emeli-dral:
Hi @SangamSwadiK ,

Thank you🔥
Could you please update the default value of ```base``` parameter to 'e', so that all our default values for drift detection stay relevant?

PR #535, opened 2023-03-21
Remove unnecessary loop in PSI calculation

Duncan-Hunter:
Hi,

I think you can remove the entire sub function and enumerate and just use numpy vectorisation here.

I haven't tested it, so I would do that first, but should make things a bit quicker and easier.

Thanks, and keep up the good work.


emeli-dral:
Hey @Duncan-Hunter ,
Thank you!

PR #537, opened 2023-03-22
enable small_distribution field for columndatadriftmetrics

mike0sv:


emeli-dral:
@mike0sv Thanks a lot!

PR #558, opened 2023-04-10
Add embeddings drift metric

mike0sv:


src/evidently/pipeline/column_mapping.py
```diff
@@ -27,6 +27,7 @@ class ColumnMapping:
     task: Optional[str] = None
     pos_label: Optional[Union[str, int]] = 1
     text_features: Optional[List[str]] = None
+    embeddings: Optional[Dict[str, List[str]]] = None
```
mike0sv, line 30: You can create an alias for this type, eg `Embeddings = Dict[str, List[str]]` so it's clearer to use

src/evidently/metrics/data_drift/embeddings_drift.py
```diff
@@ -0,0 +1,113 @@
...
+
+class EmbeddingsDriftMetricResults(MetricResult):
+    class Config:
+        dict_exclude_fields = {}
+
+    embeddings_name: str
+    drift_score: float
+    drift_detected: bool
+    method_name: str
+    plot_ref: np.ndarray
```
mike0sv, line 33: consider renaming this to `current` and `reference` or at least `current_plot`

src/evidently/metrics/data_drift/embeddings_drift.py
```diff
@@ -0,0 +1,113 @@
...
+from evidently.renderers.html_widgets import plotly_figure
+from evidently.utils.visualizations import get_gaussian_kde
+from evidently.utils.visualizations import plot_contour
+
+SAMPLE_CONSTANT = 2500
+
+
+class EmbeddingsDriftMetricResults(MetricResult):
+    class Config:
+        dict_exclude_fields = {}
```
mike0sv, line 27: probably should look like 
```
        dict_exclude_fields = {"plot_ref", "plot_cur",}
        field_tags = {k: {IncludeTags.Render} for k in dict_exclude_fields}
```

mike0sv: (I assume it's render data since it's named plot)

0lgaF: There is no IncludeTags in this branch. We could add this after merge?

Issue #552, opened 2023-04-05
Add a How-to example of using Column Mapping


elenasamuylova:
**Issue Description:** 

Evidently has a column mapping functionality that helps correctly process the input data.

It is documented here https://docs.evidentlyai.com/user-guide/input-data/column-mapping. But there is no "How to use column mapping" notebook example.

**We need to add:**

Create a "How to use column mapping" notebook similar to others (https://github.com/evidentlyai/evidently/tree/main/examples/how_to_questions)

It should include the example of using all available column mapping options for different input data. 

1. Show how to use primary mapping options (target, prediction, ID, datetime*, numerical and categorical features, text features**). 

2. Show how to use additional mapping options (datetime features*, task parameter for target function)

3. Show the different ways to pass labels and predictions for probabilistic multi-class classification. 

To show all 3 we need at least two datasets, for regression and multi-class probabilistic classification tasks. Here is an example notebook that first includes steps to download multiple datasets: https://github.com/evidentlyai/evidently/blob/main/examples/sample_notebooks/evidently_metric_presets.ipynb

*To show how to pass DateTime columns as index and features, we need to include an open example dataset (e.g., openml, creative commons public domain) that contains a DateTime column, or a step to add it artificially. 

**Text features are optional, as there is a separate How-to example for text data. 
https://github.com/evidentlyai/evidently/blob/main/examples/how_to_questions/how_to_run_calculations_over_text_data.ipynb

elenasamuylova:
Closed as completed. Thanks to @AntonisCSt for contributing!

PR #562, opened 2023-04-11
added how to use column mapping notebook

AntonisCSt:
Added most of the examples like Elena asked in the **Issue : #552** and the docs mention in: https://docs.evidentlyai.com/user-guide/input-data/column-mapping#prediction-column-s-in-classification

The notebook misses the:
* Text case
* Binary probabilistic classification, option 1
* Binary probabilistic classification, option 2

If you want I can include such examples

Also, I can cut some already existing examples to make the notebook lighter and smaller.


AntonisCSt:
Hi @emeli-dral thanks so much for the feedback. Already started adding everything you suggested and be happy to include text and embedding examples.

I just have a question:

Shouldn't the example code documentation( [Column Map](https://docs.evidentlyai.com/user-guide/input-data/column-mapping) ) in  **Binary probabilistic classification, option 1** and **Binary probabilistic classification, option 2** be

`column_mapping.pos_label = 'churn'` instead of `pos_label = 'churn'`  ?

elenasamuylova:
Hi @AntonisCSt, thanks for noticing! You are correct; we'll fix it in the docs too.

emeli-dral:
Hey @AntonisCSt ,
While reviewing the example, we found an internal issue with ```target_names```  usage in ```ClassificationConfusionMatrix```.  Here is the [link](https://colab.research.google.com/drive/1sdAz9wAyI7YmY6QTYHVscibaPRVuRxOY?authuser=1#scrollTo=GxFF_eGar-7v) in case you want to see what is broken. 
We need to make a minor fix before merging this PR to ensure the example works. We hope to resolve it in a few days.

AntonisCSt:
I see, thanks for letting me know!

I had the impression that I ran everything successfully before committing.😅

PR #615, opened 2023-05-28
spelling fix in embeddings_drift.py

ketangangal:
Fixed Spelling Error in Embeddings_drift file at line 61 & 62

From  `raise ValueError("Embeddings shold be defined in column mapping")`

To `raise ValueError("Embeddings should be defined in column mapping")`


emeli-dral:
Hey @ketangangal ,
Thanks a lot!

PR #774, opened 2023-09-20
Bump pydantic to v2

rfsan:
- I changed all imports of `pydantic` to `pydantic.v1` as defined in the [migration guide](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features)
- I had to change the minimum version of `fastapi` from `0.98.0` to `0.100.0`
- I changed the package `fastapi-utils`, that is [unmaintained](https://github.com/yuval9313/FastApi-RESTful/issues/227), to `fastapi-restful`, that is a fork. I also had to add the dependency `typing-inspect>=0.9.0` so that `fastapi-restful` works correctly with `pydantic>2`


emeli-dral:
Hi @rfsan , thank you!

mike0sv:
hey @rfsan ! We had to revert this update because v1 models do not work properly with new fastapi yet :(
See the discussion here https://github.com/tiangolo/fastapi/discussions/9966
In the meantime we will try to add compatibility layer in #792 , so you can install pydantic2 and work with it (just without UI service)

Issue #782, opened 2023-09-28
Update the sample "Metric Presets" Notebook


elenasamuylova:
**Description:**

Update the sample notebook showing how to generate `DataDriftPreset` with `"render": {"raw_data": True}` option.

**Context:**

Evidently documentation includes a sample notebook that shows all Metric Presets for tabular data.
https://github.com/evidentlyai/evidently/blob/main/examples/sample_notebooks/evidently_metric_presets.ipynb

By default, all Evidently Reports show aggregated visuals. However, there is an option to include raw data in the Reports, generating different (non-aggregated) visualizations. 

We want to add an example of using this functionality for the Data Drift report.

**Implementation**:

Edit the notebook, adding an example using the `raw_data` option for Data Drift. https://github.com/evidentlyai/evidently/blob/main/examples/data_stories/bicycle_demand_monitoring.ipynb.

Add a new header under "How to get a Python object with Report's main data?" - for example, "How to generate a Report with raw data visualizations?"

Follow the documentation on using the raw data option https://docs.evidentlyai.com/user-guide/customization/report-data-aggregation.

anirudhsudhir:
I'd like to work on this issue. Could you assign it to me?

elenasamuylova:
Hi @anirudhsudhir, we do not assign the issues directly - feel free to send the PR, and we'll review them as they come. 

I am afraid though there is already a submitted PR for this issue.

PR #790, opened 2023-10-02
Update metric presets notebook

SangamSwadiK:
**Issue:** #782 

**What does this PR do?**
Updates presets notebook by adding an example of raw data visualizations


emeli-dral:
Hey @SangamSwadiK , thank you! 
Looks great!

PR #796, opened 2023-10-03
`name` field for snapshots

mike0sv:


src/evidently/suite/base_suite.py
```diff
@@ -482,11 +483,13 @@ class ReportBase(Display):
     # collection of all possible common options
     options: Options
     id: uuid.UUID
+    name: Optional[str] = None
     timestamp: datetime
     metadata: Dict[str, MetadataValueType] = {}
     tags: List[str] = []
 
-    def __init__(self, options: AnyOptions = None, timestamp: Optional[datetime] = None):
+    def __init__(self, options: AnyOptions = None, timestamp: Optional[datetime] = None, name: str = None):
```
mou, line 491: The type definition of `name` is different in the model and the argument. But assignment is without any conversion. Is it intended?

mike0sv: When you use `None` as default argument value it's ok not to use `Optional` in type annotation

mou: I mean why not use `Optional[str]` as type for constructor argument?

mike0sv: Because it's shorter? :)

Issue #800, opened 2023-10-04
Fix docs and example on classification column mapping


elenasamuylova:
**Description:**

There is an error on the docs page regarding column mapping for binary probabilistic classification https://docs.evidentlyai.com/user-guide/input-data/column-mapping#binary-probabilistic-classification-option-1

There is the same error for all 3 options. For example, instead of:

```
column_mapping = ColumnMapping()
column_mapping.target = 'target'
column_mapping.prediction = 'prediction'
pos_label = 1
column_mapping.target_names = ['churn', 'not_churn']
```

We should specify that `pos_label` is passed inside `column_mapping`:

```
column_mapping = ColumnMapping()
column_mapping.target = 'target'
column_mapping.prediction = 'prediction'
column_mapping.pos_label = 1
column_mapping.target_names = ['churn', 'not_churn']
```

This error duplicates in the how-to example notebook: 
https://github.com/elenasamuylova/evidently/blob/main/examples/how_to_questions/how_to_use_column_mapping.ipynb

Related: https://github.com/evidentlyai/evidently/issues/799

**Implementation**

Fix the error:
- Update the docs markdown page https://github.com/evidentlyai/evidently/blob/main/docs/book/input-data/column-mapping.md
- Update the how-to example notebook https://github.com/elenasamuylova/evidently/blob/main/examples/how_to_questions/how_to_use_column_mapping.ipynb

elenasamuylova:
Hi @ujwalkumar1995, since it is a small issue - we do not directly assign them, but you are free to submit the PR. We'll review them on a first-come basis.

ujwalkumar1995:
> Hi @ujwalkumar1995, since it is a small issue - we do not directly assign them, but you are free to submit the PR. We'll review them on a first-come basis.

Hey @elenasamuylova 
Sure. Have submitted a PR. Kindly review and let me know in case of any feedback.

PR #801, opened 2023-10-04
Fix docs and example on classification column mapping

ujwalkumar1995:
This PR fixes the issue described in #800 
Please review and let me know in case of any feedback.


emeli-dral:
Hi @ujwalkumar1995 ,
Thank you for helping us out!

I like it a lot, and here is a small comment:
1. I noticed that there  is one more place in the example where pos_label is not set properly:
Screenshot 2023-10-04 at 23 38 37

Could you please update it as well?

2. We also prefer to have ```.ipynb``` examples to be stored in the repo without cell output. Could you please remove he output from all cells before you commit the example to the repo?

ujwalkumar1995:
Hey @emeli-dral 
Thank you for the feedback.
Do we need to remove the output cell from all the ipybb files or are we looking for a particular subset of files?

emeli-dral:
Well, ideally we want to remove the output cell from all the ipynb files. Practically we start from sample_notebooks and how_to_questions :)

ujwalkumar1995:
Hey @emeli-dral 
Have made the suggested changes. Kindly review.

emeli-dral:
@ujwalkumar1995 Thanks a lot!

Issue #784, opened 2023-09-29
Change the display name for DatasetCorrelationsMetric()


elenasamuylova:
**Description:**

Change the the display name for` DatasetCorrelationsMetric()`. Instead of "Dataset Correlations Metric", it should simply have "Dataset Correlations" in the title.

<img width="1283" alt="Screenshot 2023-09-29 at 03 52 04" src="https://github.com/evidentlyai/evidently/assets/67064421/8c2826e3-02cc-499c-ad03-b048a9aa14f0">

Dv04:
Hello, can I get assigned to this issue now? As this is for the Hacktoberfest, I will submit the PR as soon as it starts.

Dv04:
Hey @elenasamuylova, Just asking if I can be assigned or if should I create a PR.

elenasamuylova:
Hi @Dv04, we do not assign issues directly - but feel free to leave a comment here that you are working on it and send the PR; we will review them on a first-come basis!  

The screenshot you shared looks good!

Dv04:
Hey @elenasamuylova, Thanks. Yes I will be working on this issue and submit it as soon as possible.

Dv04:
Hey @elenasamuylova, in which branch do I have to merge the issue, as it is blocking me from doing it in the base branch.

Diksha-3905:
Hi, I’d love to work on this—can I take it

PR #814, opened 2023-10-14
Update dataset_correlations_metric.py

VishalKumar-S:
I have changed the display name for DatasetCorrelationsMetric(). Instead of "Dataset Correlations Metric", it should be "Dataset Correlations" in the title.


emeli-dral:
Hi @VishalKumar-S ,
Thank you for your contribution! 
Could you please link the issue which your PR closes? So that it counts properly.

emeli-dral:
#784

PR #834, opened 2023-10-26
add some return type hints

jameslamb:
This PR proposes adding return type hints to a few methods that were missing them. This improves the ability of type checkers like `mypy` to catch type-checking issues like "this method can return `None` but its output is being used unconditionally as if it were not `None`".

### Question for Reviewers

These are just a few things I saw while starting to familiarize myself with this codebase. There are other methods missing return type hints... would you support additional PRs like this adding return type hints?

I saw mixed use of them throughout the project, but assume they're desirable since `mypy` is run in CI:

https://github.com/evidentlyai/evidently/blob/53bda867d57408c9dae0e77578df2c76c9091e9d/.github/workflows/main.yml#L57-L58

If this type of contribution is annoying and you aren't open to more such PRs, please do tell me and I won't propose more.

Thanks very much for your time and consideration!


emeli-dral:
Hi @jameslamb ,
Thank for helping out! This type of contribution is very welcomed although we do not get it often :) 

I see that there are some failed tests. I believe it has happened because we had issues in the main branch which have already been resolved. Could you please merge main into your branch? I think this should fix tests.

jameslamb:
> This type of contribution is very welcomed

Ok thanks! I'll be making more like it then 😁 

> Could you please merge main into your branch

Just did that and updated this branch. I just merged instead of rebasing because it looks like this project squashes commits on merge, but happy to squash/rebase here if that's what yall prefer.

PR #859, opened 2023-11-19
Add RowCountTrigger to Collector Service

magdalenakuhn17:
* talking to @elenasamuylova and following https://github.com/evidentlyai/evidently/pull/734 by @mike0sv , I've added a `RowCountTrigger`

For testing:
* go to examples/integrations/collector_service
* run `evidently ui` to run UI service
* run `python src/evidently/collector/app.py` to run collector service
* run `python example_report.py` to configure collector, workspace and report and start sending data
* after someone else has tested the logic: 1) will remove unnecessary prints from `src/evidently/collector/config.py` 2) will remove `RowsCountTrigger` from `examples/integrations/collector_service/example_report.py` and enable `IntervalTrigger` again
Screenshot 2023-11-25 at 09 52 04

Additional thoughts:
* safeguard `IntervalTrigger` and `RowsCountTrigger` to not receive negative and zero as value
* I can add the `CronTrigger` or other triggers if needed

### DIFF
--- a/src/evidently/collector/config.py
+++ b/src/evidently/collector/config.py
@@ -60,9 +60,7 @@ class RowsCountTrigger(CollectorTrigger):
 
     def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bool:
         buffer_size = storage.get_buffer_size(config.id)
-        if buffer_size > 0 and buffer_size >= self.rows_count:
-            return True
-        return False
+        return buffer_size > 0 and buffer_size >= self.rows_count
 
 
 class ReportConfig(Config):
--- a/src/evidently/collector/config.py
+++ b/src/evidently/collector/config.py
@@ -63,9 +63,8 @@ class RowsCountTrigger(CollectorTrigger):
         print("RowsCountTrigger")
         print(f"rows_count: {self.rows_count}")
         buffer_size = storage.get_buffer_size(config.id)
-        print(f"buffer length:{buffer_size}")
-        if buffer_size > 0: 
-            if buffer_size % self.rows_count == 0:
+        print(f"buffer length: {buffer_size}")
+        if buffer_size > 0 and buffer_size >= self.rows_count:
                 print("Creating snapshot")
                 return True
         print("Not creating snapshot")
--- a/src/evidently/collector/config.py
+++ b/src/evidently/collector/config.py
@@ -62,9 +62,10 @@ class RowsCountTrigger(CollectorTrigger):
         print("---")
         print("RowsCountTrigger")
         print(f"rows_count: {self.rows_count}")
-        print(f"buffer length:{len(storage._buffers[config.id])}")
-        if len(storage._buffers[config.id]) > 0: 
-            if len(storage._buffers[config.id]) % self.rows_count == 0:
+        buffer_size = storage.get_buffer_size(config.id)
+        print(f"buffer length:{buffer_size}")
+        if buffer_size > 0: 
+            if buffer_size % self.rows_count == 0:
                 print("Creating snapshot")
                 return True
         print("Not creating snapshot")
--- a/src/evidently/collector/config.py
+++ b/src/evidently/collector/config.py
@@ -53,6 +53,22 @@ class IntervalTrigger(CollectorTrigger):
             self.last_triggered = now
             return True
         return False
+    
+
+class RowsCountTrigger(CollectorTrigger):
+    rows_count: int = 1
+
+    def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bool:
+        print("---")
+        print("RowsCountTrigger")
+        print(f"rows_count: {self.rows_count}")
+        print(f"buffer length:{len(storage._buffers[config.id])}")
+        if len(storage._buffers[config.id]) > 0: 
+            if len(storage._buffers[config.id]) % self.rows_count == 0:
+                print("Creating snapshot")
+                return True
+        print("Not creating snapshot")
+        return False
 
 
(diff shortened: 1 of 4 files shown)


magdalenakuhn17:
Despite formatting on my local machine with `black -l 120 -t py37 collector_service` and committing in 12ca5d150539f621769745a8077a03c5112edd4a, black executed in the github workflow still raises an error https://github.com/evidentlyai/evidently/actions/runs/6921799242/job/18827823124#step:5:28 Does anyone know how I can mimic the exact github workflow env on my local machine?

emeli-dral:
Hi @magdalenakuhn17 ,
first of all thank you so much! 
I really like what you did in this PR.

I checked the output of the workflow and it looks like there are some extra whitespaces in the  ```src/evidently/collector/config.py```

I think the best way to reproduce linter workflow run on the local machine is to run command  ```black --check --diff -l 120 -t py37 src``` or ```black --check --diff -l 120 -t py37 src/evidently/collector```  so that you check all needed files.

src/evidently/collector/config.py
```diff
@@ -53,6 +53,22 @@ def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bo
...
+
+class RowsCountTrigger(CollectorTrigger):
+    rows_count: int = 1
+
+    def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bool:
+        print("---")
+        print("RowsCountTrigger")
+        print(f"rows_count: {self.rows_count}")
+        print(f"buffer length:{len(storage._buffers[config.id])}")
+        if len(storage._buffers[config.id]) > 0:
```
mike0sv, line 66: `storage` might not have `_buffers` field, only `InMemoryStorage` has it, which is only an implementation (there are no other implementations, but still).
You can introduce a new abstract method to `CollectorStorage` like `def get_buffer_size(self, id)` and use it here instead (and implement for in-memory storage with `len(self._buffers[id])`)

src/evidently/collector/config.py
```diff
@@ -53,6 +53,22 @@ def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bo
...
+class RowsCountTrigger(CollectorTrigger):
+    rows_count: int = 1
+
+    def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bool:
+        print("---")
+        print("RowsCountTrigger")
+        print(f"rows_count: {self.rows_count}")
+        print(f"buffer length:{len(storage._buffers[config.id])}")
+        if len(storage._buffers[config.id]) > 0: 
+            if len(storage._buffers[config.id]) % self.rows_count == 0:
```
mike0sv, line 67: Are you sure you want to use modulo here instead of just `>=`? It means that if rows_count is 10, report will trigger only when buffer size is 10, 20, 30 etc, but not 11, 15, 27 etc
Also can merge this check with previous one via `and`.

mike0sv:
Hey @magdalenakuhn17 ! Thank you for this PR! I left some comments, and also you'll need to run linters like @emeli-dral suggested. Also, don't forget to remove commented code and debug prints once you are done (but I will remind you if you do forget 😄 )

src/evidently/collector/config.py
magdalenakuhn17: True, it's safer to implement the buffer size as abstract method to force any future storage implementation to provide it. Thanks for the hint :) Done in 78819402f55eb983751b641ac393c1478bfe8d03 and retested on my local machine.

src/evidently/collector/config.py
magdalenakuhn17: Combined both conditions and used `>=` instead of `%` in a705857e636245a7bd2fec6edbac01975c45434d

magdalenakuhn17:
> Hi @magdalenakuhn17 , first of all thank you so much! I really like what you did in this PR.
> 
> I checked the output of the workflow and it looks like there are some extra whitespaces in the `src/evidently/collector/config.py`
> 
> I think the best way to reproduce linter workflow run on the local machine is to run command `black --check --diff -l 120 -t py37 src` or `black --check --diff -l 120 -t py37 src/evidently/collector` so that you check all needed files.

True, I only executed black on `examples/integrations/collector_service/` :D Thanks!

src/evidently/collector/config.py
```diff
@@ -55,6 +55,16 @@ def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bo
         return False
 
 
+class RowsCountTrigger(CollectorTrigger):
+    rows_count: int = 1
+
+    def is_ready(self, config: "CollectorConfig", storage: "CollectorStorage") -> bool:
+        buffer_size = storage.get_buffer_size(config.id)
+        if buffer_size > 0 and buffer_size >= self.rows_count:
```
mike0sv, line 63: You can just return the result of this check instead of literals

mike0sv (approved):

src/evidently/collector/config.py
magdalenakuhn17: I'm not sure if I understand what you are pointing to, can you explain? Is it that you wouldn't return literal True/False, but instead the integer difference between buffer_size and rows_count or sth. else? 🤔

mike0sv: just `return buffer_size > 0 and buffer_size >= self.rows_count`

magdalenakuhn17: alright, adjusted in 9cbbe9ecd5f62ae94d4dcece22cc7ed01c781f3e

mike0sv (approved):

Issue #862, opened 2023-11-22
Bump pydantic and pyarrow version


Atharva-Phatak:
I was trying to install evidently with the latest versions of pydantic (v2) and pyarrow. I found we are still using version 11dev of pyarrow and currently pyarrow is on version 14. I think we can bump up both the dependencies and see massive perfomance gains.

rxm7706:
@Atharva-Phatak 
came here to say the same thing - but focused on pyarrow - because we ran into https://nvd.nist.gov/vuln/detail/CVE-2023-47248 - and while trying to upgrade to pyarrow 14.0.1 - evidently was a blocker. 

Also in case this helps others - since the CVE affects all PyArrow versions from 0.14.0 to 14.0.0
https://github.com/pitrou/pyarrow-hotfix provides a good work around for both pip and conda users to keep greater pyarrow support (not needing to pin to only pyarrow >14.0.1)

pip install pyarrow_hotfix
conda install -c conda-forge pyarrow-hotfix

It probably makes sense to release a pyarrow bump and version upgrade separately as Pyarrow upgrade will probably give more returns in terms of performance - and address the open security CVE.

On Pydantic - Just one note

pydantic and fastapi are interdependent - with some complexities 
See https://github.com/conda-forge/fastapi-feedstock/blob/main/recipe/meta.yaml#L26C1-L26C68
    - pydantic >=1.7.4,!=1.8,!=1.8.1,!=2.0.0,!=2.0.1,!=2.1.0,<3.0.0

1. supporting only pydantic >2 will affect dependencies and break integrations that are still pinned to pydantic <2
2. supporting both will be ideal - but was a bit extra work (based on our experience)

thank you and regards

mike0sv:
We removed pyarrow dependency in #885 and replaced it with pandas[parquet] - which means you can use any pyarrow version pandas is comfortable with.
Regarding pydantic, migration to v2 is not in our current plans. However, we did #792 so that you can use evidently with pydantic v2 (evidently models will still use pydantic.v1 classes)

dylangreenleaf:
Hoping to get some guidance on how I can use evidently with pydantic v2. The `install_requires` block within `setup.py` specifies `"pydantic<2"` which seems problematic, but perhaps I am missing something. When trying to add evidently as a dependency to my existing poetry project I get the following error:

```
➜  poetry add evidently  
Using version ^0.4.13 for evidently

Updating dependencies
Resolving dependencies... (0.1s)

[...]

PR #881, opened 2023-12-05
PyArrow upgrade to 14.0.1

cshaley:
Issue: #862 

Notes:
- upgraded pyarrow to 14.0.1
- added fsspec because 2 unit tests did not work without it
- increased evidently version so we get a new version published with new pyarrow.

Questions:
- do I need to increase the pyarrow version in the integrations folders' requirements.txt file? 
- do you like pinning with pyarrow<15? Or want a different pin?


emeli-dral:
Hi @cshaley ,
Thank you a lot for addressing this issue and making the contribution!

I apologise for the late reply. We discussed internally how to address this issue in a best way, and we were choosing between the approach you suggested and getting rid of pyarrow as dependency.

At this point, we decided to remove pyarrow from the dependencies and because of this we won't merge this PR.

I hope it solves the issue for you (the update is available in Evidently 0.4.11)

Issue #508, opened 2023-02-14
Why does TestRocAuc take so long?


ozeshun:
Hello, I am using `evidently` on developing MLOps base, and facing the issue of slow execution time of `TestRocAuc` (The data inputting for TestRocAuc has about 1 million rows).
Could you please give me some solutions?

What I verified
- I set `TestRocAuc` thresholds, but execution time did not improve [reference](https://github.com/evidentlyai/evidently/blob/main/src/evidently/tests/classification_performance_tests.py#L249-L259).
- There is no parameter in [QualityMetricsOptions](https://github.com/evidentlyai/evidently/blob/main/src/evidently/options/quality_metrics.py#L12) that improves execution time.
- It looks like `scikit-learn.metrics.roc_curve` is used when calculating AUC (Calculations using scikit-learn API should be fast) [reference](https://github.com/evidentlyai/evidently/blob/main/src/evidently/metrics/classification_performance/roc_curve_metric.py#L48-L75).

feldlime:
I faced the same problem and did some investigations.

Calculation of roc-auc invokes 6 calls of the `calculation` method of different classes. One of these classes is `ClassificationQualityMetric` which calls `evidently.calculations.classification_performance.calculate_metrics` function. This function calculates all basic classification metrics. 

First point here that you don't actually need all classification metrics to get `roc-auc` value.

Second, and the most important point is that the code is very inefficient in [these lines](https://github.com/evidentlyai/evidently/blob/main/src/evidently/calculations/classification_performance.py#L395-L402) where for every predicted value some calculations are made using `pandas`.

elenasamuylova:
Hi @feldlime, thank you for taking a look! Would you want to contribute an optimization?

feldlime:
Hi @elenasamuylova, I can try :)

PR #940, opened 2024-01-08
Optimize clf metrics calculation

feldlime:
Significantly speeded up the calculations in last `if` block in `calculate_metrics` function.

E.g. for 100k rows of random predictions it took 3 mins, now it takes 50 ms (the more data there is, the greater the difference)

Fixes https://github.com/evidentlyai/evidently/issues/508


feldlime:
I took another look and simplified the code even more

emeli-dral:
Hey @feldlime ,
many thanks!

This is a painful issue, we appreciate you addressed it. We will review it asap.

Issue #1006, opened 2024-02-23
NDCG Error for scoring 


oriol-guitart-edo:
I have an issue when calculating the NDCG for an scoring model. Attached the data to reproduce it

My code is:

```
report_test = Report(metrics=[
    PrecisionTopKMetric(k=5),
    RecallTopKMetric(k=5),
    FBetaTopKMetric(k=5),
    MAPKMetric(k=5),
    NDCGKMetric(k=5)])

column_mapping = ColumnMapping(recommendations_type='score', target='BEST_ACTION', prediction='RELEVANCE', user_id='SEARCH_ID', item_id='DEDUPID', datetime='DATE_OCCURENCE')
report_test.run(
    reference_data=None,
    current_data=request_data,
    column_mapping=column_mapping,
)

report_test.show(mode='inline')
```
[data_scores.csv](https://github.com/evidentlyai/evidently/files/14383618/data_scores.csv)

The error message:
`Exception: Dot product shape mismatch, (1,) vs (2,)`

If I remove the NDCG metric the rest of metrics work fine

PR #1007, opened 2024-02-23
Fix issue in calculating NDCG when input data is smaller than K

oriol-guitart-edo:
This PR tries to fix the [open issue](https://github.com/evidentlyai/evidently/issues/1006) which appears when some users have less recommendations than the K


emeli-dral:
Hi @oriol-guitart-edo ,
Thank you for your contribution! 

I believe reasons of failed tests are formatting and a missing bracket.
I suggest you try to fix the 55 line in the file like that: ```idcg = df.groupby("users").target.apply(lambda x: x[:i].dot(discount[: len(x)])).rename("idcg")``` (add and extra bracket after "len(x)]" and add a space in "[: len(x)]")

emeli-dral:
Looks like there is still an issue with black formatter.
Here is a way how you can check it locally: ```black -l 120 -t py38 ndcg_k.py```
I believe it is only a missing whitespace here: [: len(x)]

Issue #1012, opened 2024-02-28
Evidently Collector in v0.4.16


cayla-schuval-klaviyo:
When I try to run `evidently collector` from version 0.4.16, I am getting the following error:

```
ImproperlyConfiguredException: 500: Kwarg resolution ambiguity detected for the following keys: i, d. Make sure to use distinct keys for your dependencies, path parameters and
aliased parameters.
```

mike0sv:
thanks for reporting! let us take a look at this

c0t0ber:
@mike0sv 
I fix it by del query in type Annotation, i can make PR

mike0sv:
@c0t0ber yes, go ahead. Also you'll need to annotate some routes's return types

c0t0ber:
Fixes are implemented in the context of  https://github.com/evidentlyai/evidently/issues/1012

PR #1014, opened 2024-02-29
Feature/add new snapshot trigger and fix collector startup

c0t0ber:
Added:  New trigger has been added to initiate a snapshot for the collector based on time or the number of rows
fixed: Collector litestar server cant start https://github.com/evidentlyai/evidently/issues/1012
adde

Also I can't get mypy to run, I've tried Python versions 3.8 and 3.10
```
  File "mypy\main.py", line 95, in main
  File "mypy\main.py", line 174, in run_build
  File "mypy\build.py", line 187, in build
  File "mypy\build.py", line 270, in _build
  File "mypy\build.py", line 2867, in dispatch
  File "mypy\build.py", line 3251, in process_graph
  File "mypy\build.py", line 3372, in process_stale_scc
  File "mypy\build.py", line 2442, in write_cache
  File "mypy\build.py", line 1559, in write_cache
  File "mypy\nodes.py", line 386, in serialize
  File "mypy\nodes.py", line 3649, in serialize
  File "mypy\nodes.py", line 3581, in serialize
AssertionError: Definition of pydantic.types.JsonValue is unexpectedly incomplete
```

TestSuite does not get sent to the collector. After I forcibly called add_tests in TestSuite, the tests can be sent, but the collector cannot run them. I have not yet investigated the reasons
```
def setup_test_suite() -> ReportConfig:
    report = TestSuite(tests=[TestNumberOfOutRangeValues(column_name="values1", left=5), TestValueRange(column_name="values1")], tags=["quality"])
    rep_c =  ReportConfig.from_test_suite(report)
    print(rep_c.dict())
    return rep_c
setup_test_suite()

{'metrics': [], 'tests': [], 'options': {'color': None, 'render': None, 'custom': {}}, 'metadata': {}, 'tags': ['quality']}
```

Automatic generation of the Linestar API specification also does not work.

### DIFF
--- a/src/evidently/collector/app.py
+++ b/src/evidently/collector/app.py
@@ -1,5 +1,9 @@
 import asyncio
-from asyncio import ensure_future
+from asyncio import Task
+from typing import Any
+from typing import Awaitable
+from typing import Callable
+from typing import Dict
 from typing import List
 
 import pandas as pd
@@ -29,7 +33,6 @@ from evidently.collector.storage import LogEvent
 from evidently.telemetry import DO_NOT_TRACK_ENV
 from evidently.telemetry import event_logger
 from evidently.ui.config import NoSecurityConfig
-from evidently.ui.errors import EvidentlyServiceError
 from evidently.ui.security.no_security import NoSecurityService
 from evidently.ui.security.service import SecurityService
 from evidently.ui.security.token import TokenSecurity
@@ -38,13 +41,9 @@ from evidently.ui.security.token import TokenSecurityConfig
 COLLECTOR_INTERFACE = "collector"
 
 
-async def entity_not_found_exception_handler(request: Request, exc: EvidentlyServiceError):
-    return exc.to_response()
-
-
 @post("/{id:str}")
 async def create_collector(
-    id: Annotated[str, Parameter(query="id", description="Collector ID")],
+    id: Annotated[str, Parameter(description="Collector ID")],
     data: CollectorConfig,
     service: CollectorServiceConfig,
     storage: CollectorStorage,
@@ -58,7 +57,7 @@ async def create_collector(
 
 @get("/{id:str}")
 async def get_collector(
-    id: Annotated[str, Parameter(query="id", description="Collector ID")],
+    id: Annotated[str, Parameter(description="Collector ID")],
     service: CollectorServiceConfig,
 ) -> CollectorConfig:
     if id not in service.collectors:
@@ -68,10 +67,10 @@ async def get_collector(
 
 @post("/{id:str}/reference")
 async def reference(
-    id: Annotated[str, Parameter(query="id", description="Collector ID")],
+    id: Annotated[str, Parameter(description="Collector ID")],
     request: Request,
     service: CollectorServiceConfig,
-):
+) -> Dict[str, str]:
     if id not in service.collectors:
         raise HTTPException(status_code=404, detail=f"Collector config with id '{id}' not found")
     collector = service.collectors[id]
@@ -85,11 +84,11 @@ async def reference(
 
 @post("/{id:str}/data")
 async def data(
-    id: Annotated[str, Parameter(query="id", description="Collector ID")],
+    id: Annotated[str, Parameter(description="Collector ID")],
     request: Request,
     service: CollectorServiceConfig,
     storage: CollectorStorage,
-):
+) -> Dict[str, str]:
     if id not in service.collectors:
         raise HTTPException(status_code=404, detail=f"Collector config with id '{id}' not found")
     async with storage.lock(id):
@@ -99,7 +98,7 @@ async def data(
 
 @get("/{id:str}/logs")
 async def get_logs(
-    id: Annotated[str, Parameter(query="id", description="Collector ID")],
+    id: Annotated[str, Parameter(description="Collector ID")],
     service: CollectorServiceConfig,
     storage: CollectorStorage,
 ) -> List[LogEvent]:
@@ -108,8 +107,10 @@ async def get_logs(
     return storage.get_logs(id)
 
 
-def check_snapshots_factory(service: CollectorServiceConfig, storage: CollectorStorage):
-    async def check_snapshots():
+def check_snapshots_factory(
+    service: CollectorServiceConfig, storage: CollectorStorage
+) -> Callable[[], Awaitable[None]]:
+    async def check_snapshots() -> None:
         for _, collector in service.collectors.items():
             if not collector.trigger.is_ready(collector, storage):
                 continue
@@ -119,7 +120,7 @@ def check_snapshots_factory(service: CollectorServiceConfig, storage: CollectorS
     return check_snapshots
 
 
-async def create_snapshot(collector: CollectorConfig, storage: CollectorStorage):
+async def create_snapshot(collector: CollectorConfig, storage: CollectorStorage) -> None:
     async with storage.lock(collector.id):
         current = storage.get_and_flush(collector.id)
         if current is None:
@@ -145,10 +146,13 @@ async def create_snapshot(collector: CollectorConfig, storage: CollectorStorage)
         storage.log(collector.id, LogEvent(ok=True))
 
 
-async def loop(seconds: int, func):
-    while True:
-        await func()
-        await asyncio.sleep(seconds)
+async def loop(seconds: int, func: Callable[[], Awaitable[Any]]) -> Task:
+    async def _loop():
+        while True:
+            await func()
+            await asyncio.sleep(seconds)
+
+    return asyncio.create_task(_loop())
 
 
 def run(host: str = "0.0.0.0", port: int = 8001, config_path: str = CONFIG_PATH, secret: str = None):
@@ -161,7 +165,6 @@ def run(host: str = "0.0.0.0", port: int = 8001, config_path: str = CONFIG_PATH,
         print("Anonimous usage reporting is disabled")
     event_logger.send_event(COLLECTOR_INTERFACE, "startup")
 
-    ensure_future(loop(seconds=service.check_interval, func=check_snapshots_factory(service, service.storage)))
     security: SecurityService
     if secret is None:
         security = NoSecurityService(NoSecurityConfig())
@@ -190,6 +193,10 @@ def run(host: str = "0.0.0.0", port: int = 8001, config_path: str = CONFIG_PATH,
         if not connection.scope["auth"]["authenticated"]:
             raise NotAuthorizedException()
 
+    async def on_startup(app: Litestar) -> None:
+        # TODO: check task health
+        await loop(seconds=service.check_interval, func=check_snapshots_factory(service, service.storage))
+
     app = Litestar(
         route_handlers=[
             create_collector,
@@ -205,8 +212,8 @@ def run(host: str = "0.0.0.0", port: int = 8001, config_path: str = CONFIG_PATH,
         },
         middleware=[auth_middleware_factory],
         guards=[is_authenticated],
+        on_startup=[on_startup],
     )
-
     uvicorn.run(app, host=host, port=port)
 
 
--- a/tests/collector/test_config.py
+++ b/tests/collector/test_config.py
@@ -1,0 +1,57 @@
+import time
+from typing import Union
+from unittest.mock import Mock
+
+import pytest
+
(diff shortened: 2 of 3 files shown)


src/evidently/collector/app.py
```diff
@@ -190,6 +193,10 @@ def is_authenticated(connection: ASGIConnection, _: BaseRouteHandler) -> None:
         if not connection.scope["auth"]["authenticated"]:
             raise NotAuthorizedException()
 
+    async def on_startup(app: Litestar) -> None:
+        # TODO: check task health
+        await loop(seconds=service.check_interval, func=check_snapshots_factory(service, service.storage))
+
```
provinzkraut, line 199: The canonical way of handling long-running tasks in Litestar is to make use of the `lifespan` context managers, which are purpose built for this:

https://docs.litestar.dev/latest/usage/applications.html#lifespan-context-managers

You could define one like so:
```suggestion
@contextlib.asynccontextmanager
def lifespan(app: Litestar) -> AsyncGenerator[None, None]:
    func = check_snapshots_factory(service, service.storage)
    event = asyncio.Event()
    
    async def _loop(app: Litestar):
        while not event.is_set():
            await func()
            await asyncio.sleep(service.check_interval)

    task = asyncio.create_task(_loop())
    try:
        yield 
    finally:
          event.set()
          await task
```

and add it to `Litestar(..., lifespan=[lifespan])`.

Litestar guarantees that things you call there won't be cancelled and will gracefully shutdown, even in case of a server error. The `on_startup` hook, while it can be used to achieve similar things, doesn't explicitly guarantee this and doesn't offer a way to handle graceful shutdowns of the tasks because it's a one-off function, which means you'll end up cancelling your tasks, which can lead to some warnings or potential errors because of unfinished coroutines. It's not a big deal in this case but I thought I'd mention it :)

Alternatively, if you don't want to deal with managing the task manually at all:

```python
@contextlib.asynccontextmanager
def lifespan(app: Litestar) -> AsyncGenerator[None, None]:
    func = check_snapshots_factory(service, service.storage)

    async def _loop():
        while True:
            await func()
            await asyncio.sleep(service.check_interval)

    async with anyio.create_task_group() as tg:
      try:
          tg.start_soon(loop)
          yield 
      finally:
          tg.cancel_scope.cancel()
```

mike0sv:
As for mypy, please ensure that you are using correct version from `requirements.dev.txt`. I. personally run it in 3.10 env.
As for api spec, there is a problem with Snapshot type hints i'm investigating. For now you can copy the approach from here https://github.com/evidentlyai/evidently/blob/e27e366f61c1a4e178482d658a0c84f4f7096dcc/src/evidently/ui/api/projects.py#L286, replace snapshot model with request and parse it manually

provinzkraut:
> As for api spec, there is a problem with Snapshot type hints i'm investigating.

What is the issue exactly?

Also, just a suggestion, but if you need the raw body you can inject it via the `body` kwarg (which also then allows you to just make a sync endpoint :slightly_smiling_face:)

```python
@post("/{project_id:uuid}/snapshots")
async def add_snapshot(
    project_id: Annotated[uuid.UUID, Parameter(title="id of project")],
    body: bytes,
    project_manager: Annotated[ProjectManager, Dependency()],
    log_event: Annotated[Callable, Dependency()],
    user_id: Annotated[UserID, Dependency()],
) -> None:
    snapshot = Snapshot.parse_raw(body)
```

mike0sv:
@provinzkraut for now I know this. `Snapshot` model has `Metric` model as nested field somewhere (they are both pydantic).  And `Metric` model has private field `_context: "Context"`. It is annotated via string because of circular dependencies. Normally pydantic works with this because it ignores private fields. But when litestar tries to build schema, it tries to resolve it as ForwardRef and fails with NameError for "Context" (or something like that). I'll investigate further and probably can come up with small reproducible example

mike0sv:
created an issue https://github.com/litestar-org/litestar/issues/3150

provinzkraut:
> created an issue [litestar-org/litestar#3150](https://github.com/litestar-org/litestar/issues/3150)

Thanks, I'll look into it!

provinzkraut:
> @provinzkraut for now I know this. `Snapshot` model has `Metric` model as nested field somewhere (they are both pydantic). And `Metric` model has private field `_context: "Context"`. It is annotated via string because of circular dependencies. Normally pydantic works with this because it ignores private fields. But when litestar tries to build schema, it tries to resolve it as ForwardRef and fails with NameError for "Context" (or something like that). I'll investigate further and probably can come up with small reproducible example

This has been fixed [here](https://github.com/litestar-org/litestar/issues/3150). A patch release will be published soon!

mike0sv:
@c0t0ber schema generation should be fixed with next litestar release, cudos to @provinzkraut

src/evidently/collector/app.py
```diff
@@ -107,54 +109,45 @@ async def get_logs(
...
+    for _, collector in service.collectors.items():
+        if not collector.trigger.is_ready(collector, storage):
+            continue
+        await create_snapshot(collector, storage)
 
 
 async def create_snapshot(collector: CollectorConfig, storage: CollectorStorage) -> None:
     async with storage.lock(collector.id):
-        current = storage.get_and_flush(collector.id)
+        current = await run_in_thread_poll_executor(storage.get_and_flush, collector.id)  # FIXME: sync function
```
provinzkraut, line 121: Since this is running in a Litestar application, there's a slight benefit of having this function use the same executor as Litestar does (the executor will be re-used by Litestar), so you could just make use of `litestar.concurrency.sync_to_thread`:

```suggestion
        current = await sync_to_thread(storage.get_and_flush, collector.id)  # FIXME: sync function
```

Goes for the other instances as well.

c0t0ber:
@mike0sv 
> As for mypy, please ensure that you are using correct version from requirements.dev.txt. I. personally run it in 3.10 env

I tried running it on Windows and Linux on versions 3.10, 3.8 and I am getting this error. I perform the dependency installation using pip install -e ".[dev]" and pip install -r requirements.dev.txt according to the contributor guide
I also asked a friend to follow the steps from the contribution guide, and he encountered the same error.

provinzkraut:
@c0t0ber This seems to be a mypy bug. The version used here is a bit outdated. Upgrading to 1.8.0 solved this issue for me.

mike0sv:
We'll probably need to add tests to collector as well at some point 😅

c0t0ber:
I'll try to add smock tests

src/evidently/ui/utils.py
```diff
@@ -35,9 +35,10 @@ def _request(
             method, urllib.parse.urljoin(self.base_url, path), params=query_params, data=data, headers=headers
         )
         response.raise_for_status()
+        response_json = response.json()
```
mike0sv, line 38: this will fail if response content is not json. if you want raw dict returned, just use `dict` as response_model

mike0sv: or even better if you use actual model of the response :)

tests/collector/test_app.py
```diff
@@ -0,0 +1,98 @@
...
+
+COLLECTOR_ID = "1"
+PROJECT_ID = "2"
+
+HOST = "localhost"
+PORT = 8080
+BASE_URL = f"http://{HOST}:{PORT}"
+
+
+def create_test_data() -> pd.DataFrame:
```
mike0sv, line 25: can be fixtures

src/evidently/ui/utils.py
c0t0ber: If a generic client is really needed, then it should contain only the logic for executing the request and handling retryable exceptions, while the response processing should be implemented on the side of the specific client. The current implementation, when fully working with mypy, will only cause problems, as the function returns either T or Response. I will refactor the code, but the client needs a redesign

tests/collector/test_app.py
c0t0ber: I don't use fixtures in this case because using functions is more flexible, as it allows for the use of parameters. If a slightly different type of data is needed in the future, there will be no need to create test_data_2 and all the initialization logic is in 1 function.

mike0sv:
For some reason test for win with py3.10 fails. I restarted it multiple times in case its a timeout problem, but error persists. Do you have access to win machine to check it out?

tests/collector/test_app.py
```diff
@@ -0,0 +1,98 @@
...
+@pytest.fixture
+def server() -> None:
+    proc = Process(target=run, args=(HOST, PORT), daemon=True)
+    proc.start()
+    wait_server_start()
+    try:
+        yield
+    except Exception as e:
+        proc.kill()
+        raise e
```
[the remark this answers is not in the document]
mike0sv: let's try it, if it fails we'll yoink code from your example :)

c0t0ber:
I found an issue with passing nan, inf, -inf values; msgpack cannot handle these values and fails on JSON parsing.

One of the solutions I came up with is to parse the request manually in all places where these values could potentially appear.

@provinzkraut 
Is it possible to add a custom JSON decoder to litestar?

provinzkraut:
> Is it possible to add a custom JSON decoder to litestar?

Sure! You can extend the JSON parsing/serialisation by passing a custom encoding function for a type to the `type_decoders` available on all the app layers. For example:

```python
def is_nan(v: Any) -> bool:
 ...

def decode_nan(value: Any) -> nan:
 ...

app = Litestar(
  type_decoders=[
    (is_nan, decode_nan)
 ]
)
```

That being said, msgspec does [support](https://jcristharif.com/msgspec/supported-types.html#float) support these values. If you're encoding them to their string representation (e.g. `"nan"`), it should also be able to decode them.

tests/collector/test_config.py
```diff
@@ -0,0 +1,57 @@
...
+    assert trigger.is_ready(ready_arg, ready_arg)
+
+
+@pytest.mark.parametrize("interval", [0.0, -1.0])
+def test_interval_trigger_invalid_interval(interval: float):
+    with pytest.raises(ValidationError):
+        IntervalTrigger(interval=interval)
+
+
+def test_row_count_trigger_work():
```
mike0sv, line 30:
```suggestion
@pytest.mark.skip
def test_row_count_trigger_work():
```

tests/collector/test_app.py
```diff
@@ -0,0 +1,97 @@
...
+    proc = Process(target=run, args=(HOST, PORT))
+    proc.start()
+    wait_server_start()
+    try:
+        yield
+    finally:
+        proc.kill()
+
+
+def test_create_collector_handler_work_with_collector_client(server: None):
```
mike0sv, line 78:
```suggestion
@pytest.mark.skip
def test_create_collector_handler_work_with_collector_client(server: None):
```

tests/collector/test_app.py
```diff
@@ -0,0 +1,97 @@
...
+    collector_config = create_collector_config(create_test_data(), create_test_suite())
+    client = create_collector_client()
+    resp = client.create_collector(COLLECTOR_ID, collector_config)
+    assert resp["id"] == COLLECTOR_ID
+    assert resp["project_id"] == PROJECT_ID
+
+
+def test_data_handler_work_with_collector_client(server: None):
```
mike0sv, line 86:
```suggestion
@pytest.mark.skip
def test_data_handler_work_with_collector_client(server: None):
```

tests/collector/test_app.py
```diff
@@ -0,0 +1,97 @@
...
+    collector_config = create_collector_config(create_test_data(), create_test_suite())
+    client = create_collector_client()
+    client.create_collector(COLLECTOR_ID, collector_config)
+    client.send_data(COLLECTOR_ID, create_test_data())
+
+
+def test_references_handler_work_with_collector_client(server: None):
```
mike0sv, line 93:
```suggestion
@pytest.mark.skip
def test_references_handler_work_with_collector_client(server: None):
```

mike0sv:
Hey @c0t0ber! We decided to skip those tests for now because we got other complaints about errors in collector, so we need to your fixes merged before next release :) We can work on tests in another PR if you like, and this one we will merge once CI passes

mike0sv:
As for nan/inf issue, can you give us a reproducible example?

c0t0ber:
@mike0sv 

> As for nan/inf issue, can you give us a reproducible example?

example here https://github.com/evidentlyai/evidently/issues/1031

PR #1045, opened 2024-03-28
Assorted refactorings of the collector and UI app

provinzkraut:
Hi all, this is a bunch of updates pertaining various things introduced in #995, when moving from FastAPI to Litestar.

Some of them are simple refactorings to simplify code and using Litestar best-practices, others should provide a real world performance boost. 

I'll try to add comments to the various changes to explain why they have been made.

There's still some things that could be improved, but given my lack of familiarity with the codebase, I didn't want to go ahead and change too many things in code that's (as far as I can tell?) not (yet?) covered by unit tests.


setup.py
```diff
@@ -62,7 +62,7 @@
         "pydantic>=1.10.14",
         "litestar>=2.6.3",
         "typing-inspect>=0.9.0",
-        "uvicorn>=0.22.0",
+        "uvicorn[standard]>=0.22.0",
```
provinzkraut, line 65: This includes some dependencies which can greatly improve uvicorn's performance, such as uvloop and `httptools` instead of `h11`

src/evidently/collector/app.py
```diff
@@ -46,14 +46,12 @@
 logger = logging.getLogger(__name__)
 
 
-@post(
-    "/{id:str}",
-)
-async def create_collector(
+@post("/{id:str}", sync_to_thread=True)
+def create_collector(
```
provinzkraut, line 50: I've changed these to synchronous endpoints with `sync_to_thread=True` because they were all performing some sort of blocking operation, usually related to the `CollectorService`/`CollectorStorage`; They both perform blocking IO under the hood, which means that, when called from within an async function like this, they will block the event loop for the whole duration, effectively preventing any other requests from running concurrently. (some further reading can be done [here](https://docs.litestar.dev/latest/topics/sync-vs-async.html))

The easy fix here is to use synchronous endpoints and let Litestar handle the concurrency, but a more elegant way to solve this would be to make the underlying APIs asynchronous as well. From looking at the code briefly, this shouldn't be too much of an issue, but I didn't know if they were external dependencies on these / APIs that would break were those changes introduces.

src/evidently/collector/app.py
```diff
@@ -46,14 +46,12 @@
...
     id: Annotated[str, Parameter(description="Collector ID")],
     parsed_json: CollectorConfig,
-    service: CollectorServiceConfig,
-    storage: CollectorStorage,
+    service: Annotated[CollectorServiceConfig, Dependency(skip_validation=True)],
+    storage: Annotated[CollectorStorage, Dependency(skip_validation=True)],
```
provinzkraut, line 54: I'm not entirely sure why these are Pydantic models, as they don't seem to be serialized/validated anywhere. 

Them being Pydantic models however, and injected via Litestar's DI, means that they'll get validated on every call, which, for complex Pydantic objects, especially when using Pydantic 1, incurs a not insignificant performance overhead. Using `skip_validation=True` prevents this validation from taking place.

src/evidently/collector/app.py
```diff
@@ -89,25 +87,25 @@ async def reference(
...
 
-@post("/{id:str}/data")
-async def data(
+@post("/{id:str}/data", sync_to_thread=True)
+def data(
     id: Annotated[str, Parameter(description="Collector ID")],
-    request: Request,
-    service: CollectorServiceConfig,
-    storage: CollectorStorage,
+    data: Any,
```
provinzkraut, line 93: This achieves the same as calling `await request.json()`, but allows using a synchronous function

src/evidently/collector/app.py
```diff
@@ -89,25 +87,25 @@ async def reference(
...
+    service: Annotated[CollectorServiceConfig, Dependency(skip_validation=True)],
+    storage: Annotated[CollectorStorage, Dependency(skip_validation=True)],
 ) -> Dict[str, str]:
     if id not in service.collectors:
         raise HTTPException(status_code=404, detail=f"Collector config with id '{id}' not found")
     async with storage.lock(id):
-        storage.append(id, await request.json())
+        storage.append(id, data)
```
provinzkraut, line 100: So far I could only see an in-memory implementation of this interface, but seeing as this isn't async, it could become problematic if a blocking implementation is added. This would probably be better as a fully async service as well

src/evidently/collector/app.py
```diff
@@ -219,10 +217,10 @@ async def check_service_snapshots_periodically():
...
         ],
         dependencies={
-            "security": Provide(lambda: security, use_cache=True, sync_to_thread=True),
-            "service": Provide(lambda: service, use_cache=True, sync_to_thread=True),
-            "storage": Provide(lambda: service.storage, use_cache=True, sync_to_thread=True),
-            "parsed_json": Provide(parse_json),
+            "security": Provide(lambda: security, use_cache=True, sync_to_thread=False),
+            "service": Provide(lambda: service, use_cache=True, sync_to_thread=False),
+            "storage": Provide(lambda: service.storage, use_cache=True, sync_to_thread=False),
+            "parsed_json": Provide(parse_json, sync_to_thread=False),
```
provinzkraut, line 223: Using `sync_to_thread=True` here only adds unnecessary overhead as these function are all already non-blocking

src/evidently/ui/api/projects.py
provinzkraut, line 1: Same as for the collector app, the `ProjectManager` methods used in the endpoints are blocking, which is why I've switched them to use `sync_to_thread=True`.

src/evidently/ui/api/projects.py
```diff
@@ -28,18 +29,24 @@
...
+from evidently.ui.errors import NotEnoughPermissions
 from evidently.ui.type_aliases import OrgID
 from evidently.ui.type_aliases import TeamID
 from evidently.ui.type_aliases import UserID
 from evidently.utils import NumpyEncoder
 
 
-@get("/{project_id:uuid}/reports")
-async def list_reports(
+def is_authenticated(connection: ASGIConnection, _: BaseRouteHandler) -> None:
```
provinzkraut, line 39: Defining this guard here instead of the main app file allows for it to be used as a guard directly on the handlers. This is makes it more explicit to the reader which auth is being performed where and avoids the nested router setup and the factory pattern used before to create the routers.

src/evidently/ui/api/projects.py
```diff
@@ -28,18 +29,24 @@
...
+        raise NotEnoughPermissions()
+
+
+@get("/{project_id:uuid}/reports", sync_to_thread=True)
+def list_reports(
     project_id: Annotated[uuid.UUID, Parameter(title="id of project")],
-    project_manager: Annotated[ProjectManager, Dependency()],
-    log_event: Annotated[Callable, Dependency()],
-    user_id: Annotated[UserID, Dependency()],
+    project_manager: Annotated[ProjectManager, Dependency(skip_validation=True)],
```
provinzkraut, line 47: Same as for the collector app, validating these Pydantic models on every invokation of the endpoints might be costly.

src/evidently/ui/api/projects.py
```diff
@@ -135,14 +142,14 @@ async def list_test_suites(
...
 
-@get("/{project_id:uuid}/{snapshot_id:uuid}/graphs_data/{graph_id:str}")
-async def get_snapshot_graph_data(
+@get("/{project_id:uuid}/{snapshot_id:uuid}/graphs_data/{graph_id:str}", sync_to_thread=True)
+def get_snapshot_graph_data(
     project_id: Annotated[uuid.UUID, Parameter(title="id of project")],
     snapshot_id: Annotated[uuid.UUID, Parameter(title="id of snapshot")],
     graph_id: Annotated[str, Parameter(title="id of graph in snapshot")],
-    project_manager: Annotated[ProjectManager, Dependency()],
-    log_event: Annotated[Callable, Dependency()],
```
provinzkraut, line 144: The usage of `Dependency()` here is redundant

src/evidently/ui/api/projects.py
```diff
@@ -28,18 +29,24 @@
...
+def is_authenticated(connection: ASGIConnection, _: BaseRouteHandler) -> None:
+    if not connection.scope["auth"]["authenticated"]:
+        raise NotEnoughPermissions()
+
+
+@get("/{project_id:uuid}/reports", sync_to_thread=True)
+def list_reports(
     project_id: Annotated[uuid.UUID, Parameter(title="id of project")],
-    project_manager: Annotated[ProjectManager, Dependency()],
-    log_event: Annotated[Callable, Dependency()],
```
provinzkraut, line 41: The usage of an empty `Dependency()` is redundant (applies for the other cases in this file as well)

src/evidently/ui/api/projects.py
```diff
@@ -224,15 +231,15 @@ async def list_project_dashboard_panels(
...
 
 
-@get("/{project_id:uuid}/dashboard")
-async def project_dashboard(
+@get("/{project_id:uuid}/dashboard", sync_to_thread=True)
+def project_dashboard(
     project_id: Annotated[uuid.UUID, Parameter(title="id of project")],
     # TODO: no datetime, as it unable to validate '2023-07-09T02:03'
-    timestamp_start: Annotated[Optional[str], Parameter(default=None)],
-    timestamp_end: Annotated[Optional[str], Parameter(default=None)],
```
provinzkraut, line 232: Defaults can be set for the parameters directly

src/evidently/ui/api/static.py
```diff
@@ -1,50 +1,42 @@
...
-from litestar import Litestar
 from litestar import MediaType
-from litestar import Request
 from litestar import Response
-from litestar.static_files import StaticFilesConfig
+from litestar.response import File
+from litestar.router import Router
+from litestar.static_files import create_static_files_router
 
+base_path = pathlib.Path(__file__).parent.parent.resolve() / "ui"
```
provinzkraut, line 10: This should probably be replaced with [`importlib.resources`](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), but I wasn't sure how you're packacking this so I left it as is

src/evidently/ui/api/static.py
```diff
@@ -1,50 +1,42 @@
...
-            "/auth",
-            "/teams",
-            "/token",
-            "/projects/{path:path}",
-            "/teams/{path:path}",
-        ],
-        include_in_schema=False,
-    )
-    async def index(path: Optional[str]) -> Response:
-        data = open(pathlib.Path(ui_path) / "index.html").read()
```
provinzkraut, line 28: This was blocking IO; Explanation see above

src/evidently/ui/api/static.py
```diff
@@ -1,50 +1,42 @@
...
+        filename="index.html",
+        media_type=MediaType.HTML,
+        content_disposition_type="inline",
     )
-    async def manifest(request: Request) -> Response:
-        path = request.url.path[1:]
-        if path in ("manifest.json", "favicon.ico", "favicon-16x16.png", "favicon-32x32.png"):
-            data = os.path.join(ui_path, path)
-            with open(data, "rb") as f:
-                return Response(content=f.read())
```
provinzkraut, line 45: More blocking IO

src/evidently/ui/app.py
```diff
@@ -104,36 +95,31 @@ async def middleware(scope: Scope, receive: Receive, send: Send) -> None:
...
-            "org_id": Provide(get_org_id),
-            "log_event": Provide(get_event_logger),
-            "parsed_json": Provide(parse_json),
+            "user_id": get_user_id,
+            "org_id": get_org_id,
+            "log_event": get_event_logger,
+            "parsed_json": Provide(parse_json, sync_to_thread=False),
         },
         middleware=[auth_middleware_factory],
-        debug=True,
```
provinzkraut, line 134: Running in debug mode by default could cause some issues in production

src/evidently/ui/errors.py
```diff
@@ -1,17 +1,6 @@
-import json
-
 from litestar import Response
 
 
-class JSONResponse(Response):
```
provinzkraut, line 6: This isn't needed; The `Response` class can already deal with everything Litestar can serialise.

src/evidently/ui/utils.py
```diff
@@ -46,30 +46,5 @@ def _request(
         return response
 
 
-_skip_jsonable_encoder_cache = {}
-
-
-def skip_jsonable_encoder(f):
```
provinzkraut, line 52: This wasn't being used and seems to be a leftover from the FastAPI app

src/evidently/ui/api/projects.py
Liraim: It was mostly used as explicit indicator of parameter to be a dependency.

src/evidently/ui/ui/assets/favicon-16x16.png
DimaAmega, line 1: Hi @provinzkraut

I think we don't need this `ui/ui/assets`... just `ui/assets` instead. 
And i created [PR](https://github.com/provinzkraut/evidently/pull/1) for this issue into your fork to fix it properly

provinzkraut:
@mike0sv I've reverted the changes as discussed

DimaAmega:
@provinzkraut Let's merge [this pr](https://github.com/provinzkraut/evidently/pull/2)

Issue #1060, opened 2024-04-05
Save report to file cause "unclosed file" ResourceWarning


marviniter:
## save report with code:
```
drift_report = Report(metrics=[
DataDriftPreset(), TargetDriftPreset()
])
drift_report.run(reference_data=reference, current_data=current, column_mapping=column_mapping)
drift_html_file = "./drift_report.html"
drift_report.save_html(drift_html_file)
drift_json_file = "./drift_report.json"
drift_report.save_json(drift_json_file)
```
## Unclose file Warning log with:
```
D:\ProgramData\Anaconda3\envs\models\lib\site-packages\evidently\utils\dashboard.py:231: ResourceWarning:
unclosed file <_io.TextIOWrapper name='D:\ProgramData\Anaconda3\envs\models\lib\site-packages\evidently\nbextension\static\index.js' mode='r' encoding='utf-8'>
D:\ProgramData\Anaconda3\envs\models\lib\site-packages\evidently\utils\dashboard.py:236: ResourceWarning:
unclosed file <_io.BufferedReader name='D:\ProgramData\Anaconda3\envs\models\lib\site-packages\evidently\nbextension\static\material-ui-icons.woff2'>
```

PR #1061, opened 2024-04-05
fix dashboard unclose open file

marviniter:
fix issue:#1060 unclose file error


DimaAmega:
@marviniter Thanks a lot for this one! 😊

PR #1109, opened 2024-05-09
improve categorical features selection in data definition

svetleo-evidently:


src/evidently/utils/data_preprocessing.py
```diff
@@ -510,12 +552,23 @@ def _get_column_presence(column_name: str, data: _InputData) -> ColumnPresenceSt
...
 NUMBER_UNIQUE_AS_CATEGORICAL = 5
 
 
-def _get_column_type(column_name: str, data: _InputData, mapping: Optional[ColumnMapping] = None) -> ColumnType:
+def _get_column_type(
+    column_name: str, data: _InputData, mapping: Optional[ColumnMapping] = None, cardinality_limit: Optional[int] = None
+) -> ColumnType:
     if mapping is not None:
         if mapping.categorical_features and column_name in mapping.categorical_features:
+            if cardinality_limit and _is_cardinality_exceeded(column_name, data, cardinality_limit):
```
Liraim, line 570: Better reuse result of `_is_cardinality_exceeded` as it is expensive call.

src/evidently/utils/data_preprocessing.py
```diff
@@ -309,8 +332,27 @@ def create_data_definition(
...
+        _process_column(
+            column_name,
+            data,
+            if_partially_present="skip",
+            mapping=mapping,
+            cardinality_limit=categorical_features_cardinality_limit,
+        )
+        for column_name in data_columns
     ]
+    # for column_name in data_columns:
```
Liraim, line 344: Commented code should be deleted.

src/evidently/utils/data_preprocessing.py
svetleo-evidently: I use _is_cardinality_exceeded twice in _get_column_type but in different "if" branches, so _is_cardinality_exceeded never will be called twice.

PR #1145, opened 2024-06-06
UI~Bakend typing

DimaAmega:


src/evidently/ui/api/projects.py
```diff
@@ -207,7 +214,7 @@ def get_snapshot_data(
     project_manager: Annotated[ProjectManager, Dependency(skip_validation=True)],
     log_event: Callable,
     user_id: UserID,
-) -> Response:
+) -> str:
```
Liraim, line 217: We lose media-type in this approach, maybe need to look at `-> Response[str]` so we can customize response as was before.

src/evidently/model/widget.py
```diff
@@ -85,14 +86,18 @@ class BaseWidgetInfo:
     alertsPosition: Optional[str] = None
     alertStats: Optional[AlertStats] = None
     params: Any = None
-    insights: Iterable[Insight] = ()
-    additionalGraphs: Iterable[Union[AdditionalGraphInfo, "BaseWidgetInfo", PlotlyGraphInfo]] = ()
-    alerts: Iterable[Alert] = ()
-    tabs: Iterable["TabInfo"] = ()
-    widgets: Iterable["BaseWidgetInfo"] = ()
+    insights: Iterable[Insight] = field(default_factory=lambda: [])
```
mike0sv, line 89: `default_factory=list` is better

src/evidently/options/color_scheme.py
```diff
@@ -5,14 +5,14 @@
 
 RED = "#ed0400"
 GREY = "#4d4d4d"
-COLOR_DISCRETE_SEQUENCE = (
+COLOR_DISCRETE_SEQUENCE = [
```
mike0sv, line 8: but why

src/evidently/ui/api/models.py
```diff
@@ -37,7 +38,6 @@ def from_metric(cls, metric: Metric):
 class ReportModel(BaseModel):
     id: uuid.UUID
     timestamp: datetime.datetime
-    metrics: List[MetricModel]
```
mike0sv, line 40: we dont use it in UI?

src/evidently/ui/api/models.py
```diff
@@ -82,9 +85,10 @@ def from_snapshot(cls, snapshot: SnapshotMetadata):
         )
 
 
-class DashboardInfoModel(BaseModel):
+@dataclass
```
mike0sv, line 88: but why

src/evidently/ui/api/projects.py
```diff
@@ -156,7 +163,7 @@ def get_snapshot_graph_data(
...
```
mike0sv, line 166: why

src/evidently/ui/api/projects.py
```diff
@@ -243,6 +250,24 @@ def list_project_dashboard_panels(
     return list(project.dashboard.panels)
 
 
+# We need this endpoint to export
+# some additional models to open api schema
+@get("/models/additional")
```
mike0sv, line 255: there might be a way to do it better, I will ask smbd from litestar team

src/evidently/ui/app.py
```diff
@@ -42,9 +43,17 @@ def run_local(
     secret = secret or os.environ.get(EVIDENTLY_SECRET_ENV)
     if secret is not None:
         config.security = TokenSecurityComponent(token=SecretStr(secret))
+
+    if return_litestar_app:
```
mike0sv, line 47: i think it's better to do the other way around. eg create `configure_app` with the same signature as `run_local` that does everything except `run(config)` and make `run_local(...): run(configure_app(...))` and `litestar_app(): create_app(configure_app())`

src/evidently/ui/api/models.py
DimaAmega: yep

src/evidently/ui/api/models.py
DimaAmega: First of all we need to support `widgets` type as `List[BaseWidgetInfo]`

`BaseWidgetInfo` is dataclass and seems like we can't use dataclass as pydantic field (I've been getting errors 😝) -> the only way is to convert `DashboardInfoModel` to dataclass as well

src/evidently/ui/api/projects.py
DimaAmega: In order to get correct types in UI:
image

And this is string (not a json) because may contains NaN, +-Infinity and right now we are using special json parser

src/evidently/ui/api/models.py
mike0sv: or convert BaseWidgetInfo into BaseModel :)

src/evidently/options/color_scheme.py
DimaAmega: seems like `litestar` open api schema generation doesn't support tuples well

src/evidently/ui/app.py
DimaAmega: fixed

PR #1150, opened 2024-06-12
Support target / prediction as feature in Classification Performance.

Liraim:


svetleo-evidently:
I cheked locally, it works for me for ClassificationPreset with render.raw_data

PR #1192, opened 2024-07-08
add datasets to cloud workspace

mike0sv:


src/evidently/ui/workspace/cloud.py
```diff
@@ -120,6 +138,23 @@ def create_team(self, team: Team, org_id: OrgID = None) -> TeamModel:
             response_model=TeamModel,
         )
 
+    def add_dataset(
+        self, file: BinaryIO, name: str, org_id: OrgID, team_id: TeamID, description: Optional[str]
+    ) -> DatasetID:
+        response = self._request(
+            "/api/datasets/",
+            "POST",
+            body={"name": name, "description": description, "file": file},
```
svetleo-evidently, line 147: column_mapping missing

svetleo-evidently: I mean for UI we have a separate handler patch datasets. For sdk maybe we want to have an opportunity to set columnMapping right from add_dataset

PR #1197, opened 2024-07-09
Fix: Encode str to bytes before writing to buffered IO

mesquita:
If you try something like this:
```
html_buffer = BytesIO()
report.save_html(html_buffer)
```
where report is 
```
report = Report(
    metrics=[
        DataDriftPreset(),
    ]
)
```
after a successful `report.run`

You'll get:
```
TypeError: a bytes-like object is required, not 'str'
```
due to the fact that `render` in `filename.write(render)` is a string, not bytes. 
In this PR, I'm encoding render to bytes before writing to the buffer.

### DIFF
--- a/src/evidently/suite/base_suite.py
+++ b/src/evidently/suite/base_suite.py
@@ -1,6 +1,7 @@
 import abc
 import copy
 import dataclasses
+import io
 import json
 import logging
 import uuid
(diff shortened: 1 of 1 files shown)


mike0sv:
Can't you just use `StringIO` instead of `BytesIO`?

mesquita:
> Can't you just use `StringIO` instead of `BytesIO`?

Hi @mike0sv, since `save_html` accepts `str` or `IO`, it should work with BytesIO as well. In my case, it'd be very useful.

I can suggest something like this:

```
          if isinstance(filename, str):
              with open(filename, "w", encoding="utf-8") as out_file:
                  out_file.write(render)
          elif isinstance(filename, io.BytesIO):
              filename.write(render.encode(encoding="utf-8"))
          elif isinstance(filename, io.StringIO):
              filename.write(render)
          else:
              raise ValueError("filename should be a string or a file-like object")
```

what do you think?

mike0sv:
This will not work because of duck typing. Those instance checks will return `False` for objects created with `open(...)` for example. Same for `typing.BinaryIO` and `typing.TextIO`.
If you need to get html payload, you can use `get_html` instead of `save_html`.
As for type annotation, probably we should correct annotation to `typing.TextIO` or use more flexible check. I don't know how to do this from the top of my head though

PR #1281, opened 2024-09-03
Add API to reload state from FS. 

nikp1172:
If some live changes are made to storage backend/filesystem, the changes are not updated and the whole UI service needs to restart.

This button refreshed the whole state.

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -167,16 +167,8 @@ jobs:
         working-directory: ui
         run: pnpm dlx playwright@1.43.0 install --with-deps
 
-      - uses: actions/cache@v3
-        id: cache-bikes-dataset
-        env:
-          cache-name: cache-bikes-dataset
-        with:
-          path: Bike-Sharing-Dataset.zip
-          key: cache-bikes-dataset
-      - name: Download test data
-        if: ${{ steps.cache-bikes-dataset.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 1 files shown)


DimaAmega:
Hi @nikp1172

Have you tried using the `refresh snapshots` button inside some project?

image


Does this solve the problem? If no, please describe your use case in more details 😊

nikp1172:
@DimaAmega  It does not. It refreshes reports and test suites. But NOT the dashboard panels. They are static.

So I have a Job which updates the dashboard to s3, now reports get updated but NOT the dashboard.

Also, sometimes the job can create a project -> but a new project also doesn't show up.

 We need a global refresh button for this

DimaAmega:
> @DimaAmega It does not. It refreshes reports and test suites. But NOT the dashboard panels. They are static.
> So I have a Job which updates the dashboard to s3, now reports get updated but NOT the dashboard.
> Also, sometimes the job can create a project -> but a new project also doesn't show up.
> We need a global refresh button for this

Hey @nikp1172

We've talked about it and we think it's a nice feature to have. But we suggest you to make changes only to the backend by rolling back the UI changes, then we will make changes to the UI ourselves in the next PR immediately

I'll explain. It is quite difficult to describe the details of how we want to see the final result in the UI, most likely it will be an iterative process, but we will keep your original idea with the "reload all" button in the header

Does it sound good?

nikp1172:
Sure, @DimaAmega let me do that

nikp1172:
Created a new PR here: https://github.com/evidentlyai/evidently/pull/1294

Closed this for now

PR #1303, opened 2024-09-13
Fixing minor typo

sapenov:
Fixing some minor typos and also tweaking install to add spark

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,14 +72,14 @@ See official pip documentation for examples and explanations: https://pip.pypa.i
 
 ```sh
 cd /path/to/evidently_repo
-pip install -e ".[dev,spark]"
+pip install -e ".[dev]"
 ```
 
 #### Windows
 
 ```sh
 cd C:\path\to\evidently_repo
-pip install -e .[dev,spark]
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -72,14 +72,14 @@ See official pip documentation for examples and explanations: https://pip.pypa.i
 
 ```sh
 cd /path/to/evidently_repo
-pip install -e ".[dev]"
+pip install -e ".[dev,spark]"
```
DimaAmega, line 75: I don't think we need to have `spark` extra here. I mean, yes, we have a few extras but only `dev` is useful for like contributing

sapenov: Ok, I'll take it out. 
In my environment, the pytest fails without it.

DimaAmega: > Ok, I'll take it out. In my environment, the pytest fails without it.

Hm... Okay
I think we have to improve our docs about how to run tests :)
Thanks for info about it

DimaAmega (approved):

DimaAmega:
@sapenov Thanks for this fix! You have attentive eyes 👀😊

PR #1332, opened 2024-10-04
Add is valid sql descriptor

### LINKED ISSUE #1321: Add a new `IsValidSQL()` descriptor to Evidently
About Hacktoberfest contributions: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest-2024

**Description**

Add the `IsValidSQL()` descriptor that checks whether the text in a specified column is a valid SQL query, without executing the query. This descriptor returns True for valid SQL queries and False for invalid queries.

Sifr-un:
Implementing a SQL validator for LLM responses as a new descriptor.
Resolves #1321


DimaAmega:
Hi @Rayryu

You have some CI failed jobs, please check it

Rayryu:
> [DimaAmega](/DimaAmega)

I've pushed a typo fix.

src/evidently/features/is_valid_sql_feature.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any
+from typing import ClassVar
+from typing import Optional
+
+import sqlvalidator
```
DimaAmega, line 5: Let's move this import to the [beginning](https://github.com/evidentlyai/evidently/pull/1332/files#diff-a1bc3e164c7afaa840d5966343c65f5676709a33346a1f84166f9a0eed89f649R32) of the `is_valid_sql` function

Rayryu: Changes made. What was causing the issue?

src/evidently/features/is_valid_sql_feature.py
```diff
@@ -0,0 +1,44 @@
...
+        super().__init__()
+
+    def apply(self, value: Any):
+        if value is None or not isinstance(value, str):
+            return False
+
+        return self.is_valid_sql(value)
+
+    def is_valid_sql(self, query: str) -> bool:
+        queries = query.strip().split(";")  # Split by semicolon
```
DimaAmega, line 32: Let's consider the following sql

```sql
SELECT * FROM your_table WHERE your_column LIKE '%;%';
```

or

```sql
SELECT CONCAT('Hello', '; ', 'World') AS greeting_with_semicolon;
```

we will get invalid in your implementation but it is actually valid sql

Do we actually need to split by `;`?

Rayryu: The split by `;` is a bit of a stretch since the column could contain multiple SQL queries. However, we can stick to the basic usage and treat the column values as single queries. 
How do we view the scope of this feature?

src/evidently/features/is_valid_sql_feature.py
DimaAmega: Now all the UI checks are being passed in CI 😊

It's a bit tricky: other fix is to add `llm` extras to install step in these tests but `llm` extra at the same time is kinda overkill and unnecessary

mike0sv:
Hi! To fix mypy you need to add ignore block to setup.cfg in repo root. You can check out mypy section for examples

Rayryu:
> Hi! To fix mypy you need to add ignore block to setup.cfg in repo root. You can check out mypy section for examples

Thanks for chiming in. I've added a `sqlvalidator `block in `setup.cfg`.

emeli-dral:
Hi @Rayryu,

It looks like we're almost there!

Could you please add the new descriptor to the list of descriptors under the TextEvals section on the [documentation page](https://docs.evidentlyai.com/reference/all-metrics#text-evals-1)? The file that needs to be updated is located here: [all-metrics.md#text-evals](https://github.com/evidentlyai/evidently/blob/main/docs/book/reference/all-metrics.md#text-evals).

Once that’s done, we’ll be all set to proceed with the merge!

emeli-dral:
Hi @Sifr-un ,
do you need some help with adding ```IsValidSQL``` to the list of descriptors under the TextEvals section  [all-metrics.md#text-evals](https://github.com/evidentlyai/evidently/blob/main/docs/book/reference/all-metrics.md#text-evals)?

vectorvp:
@emeli-dral, how can I help to speed this up? Merge the changes from this branch and prepare new PR?

Issue #1309, opened 2024-09-23
Add a New `WordNoMatch` Descriptor to Evidently


elenasamuylova:
# Add a New `WordNoMatch` Descriptor to Evidently

About Hacktoberfest contributions: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest-2024

## Description:

Evidently already has an `ExcludesWords()` descriptor that checks if the text does **not** contain any specified words from a shared list. 

However, sometimes you might need to check that the text does not contain words specific to each row instead of a shared list.

### Example:

| **Question**                                 | **Response**                                                                 | **Forbidden Words**                      |
|----------------------------------------------|-------------------------------------------------------------------------------|------------------------------------------|
| "Can I cancel my subscription at any time?"  | "You are allowed to cancel at any time, and we guarantee that you will receive a refund." | ["guarantee", "allowed", "refund"]       |

## What to Implement:

The new **WordNoMatch()** descriptor should:

1. **Accept a `with_column` parameter**: This column contains a list of forbidden words specific to each row.
2. **Accept a `lemmatize` parameter** (default `True`): When `True`, this will consider inflected or variant forms of words. Works the same as in the `ExcludesWords()` descriptor.
3. **Return True/False**:
   - Return `True` if the words from the list are **not** present in the row's text.
   - Return `False` if forbidden words are present.

## References:

- Check the `ExcludesWords()` descriptor as reference.
- For a two-column descriptor implementation, check the `SemanticSimilarity` descriptor and the `CustomPairColumnEval` template.

jon-bown:
Opened a joint PR for this issue and #1308

Issue #1308, opened 2024-09-23
Add a new `WordMatch` descriptor to Evidently


elenasamuylova:
# Add a New `WordMatch` Descriptor to Evidently

About Hacktoberfest contributions: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest-2024

##  Description:

Evidently already has an `IncludesWords()` descriptor that checks if the text contains `any` (by default) or `all` specified words, returning a True/False result for each row. However, this descriptor uses a single shared list of words for all rows.

In some cases, such as when evaluating responses against specific ground truth answers, you may need a different list of words for each row. For example, you might want to check if generated responses contain the expected keywords for each row:

**Example:**

| **Question**                     | **Generated Response**                  | **Expected Words**               |
|-----------------------------------|-----------------------------------------|----------------------------------|
| "Name a primary color."           | "Red is a primary color."               | ["blue", "red", "yellow"]        |

## What to Implement:

The new **WordMatch()** descriptor should:

1. **Accept a `with_column` parameter**: This column contains a list of words specific to each row.
2. **Accept a `lemmatize` parameter**. Default `True`, to consider inflected and variant words. (Same as `IncludesWords()` descriptor).
3. **Allow configuration for `any` or `all`** words present. (Same as `IncludesWords()` descriptor):
4. **Return True/False for each row**` if the specified condition (any or all) is met or not.

**References:**
- Check the `IncludesWords` descriptor for vocabulary word check implementation.
- For a two-column descriptor implementation, check the `SemanticSimilarity` descriptor and the `CustomPairColumnEval` template.

jon-bown:
Along with the `with_column` parameter, would the column we are getting the generated response from always be named the same? Or would we need it to resemble `SemanticSimilarity` implementation with two arbitrary columns being compared  with `WordMatch`?

PR #1334, opened 2024-10-05
Add WordMatch + WordNoMatch descriptor

jon-bown:
Implement `WordMatch`, `WordNoMatch`,  descriptor + feature and associated unit tests. 

Resolves issue #1308 
Resolves issue #1309


emeli-dral:
Hi @jon-bown ,
This one looks good as well!

Could you please double check your ```src/evidently/features/_registry.py```?
It looks like paths to feature classes are invalid.

jon-bown:
> Hi @jon-bown , This one looks good as well!
> 
> Could you please double check your `src/evidently/features/_registry.py`? It looks like paths to feature classes are invalid.

Done! Also want to point out that the error you're getting with my ItemMatch PR would need to be resolved here also.

emeli-dral:
> Done! Also want to point out that the error you're getting with my ItemMatch PR would need to be resolved here also.

Thank you for the update, fair enough! I suggest you change a list to a tuple here as well, similar to what we’ve discussed in the related issue. This should help prevent the error and maintain consistency across both implementations.

Thanks for your attention to detail! 🙏

emeli-dral:
Hi @jon-bown ,
It looks like there’s a small issue with the Python f-string here:
```
File "/home/runner/work/evidently/evidently/src/evidently/features/words_feature.py", line 198
    default_display_name=f"Text contains {self.mode.split("_")[1]} defined words"
```
I believe the string is being parsed incorrectly due to the nested double quotes inside the f-string. To resolve this, try changing the internal double quotes to single quotes ('_').

That should fix the issue!

jon-bown:
Hi @emeli-dral I believe I've fixed all the issues with this PR. Can you let me know if anything else needs to change? Thanks!

emeli-dral:
Hi @jon-bown,
I had a chance to review your repo, and I really like how the descriptor is working—great job! However, I would like to request one more small change, apologies for the extra ask!

To keep things consistent and avoid potential confusion for users, I’d suggest making the WordMatch descriptor work directly with tuples, just like the ItemMatch descriptor. Having one descriptor use tuples and another use lists might be a bit unclear.

In the future, we plan to support lists as an input type in a more centralized way, and we’ll introduce list support across all relevant descriptors then.

Thanks again for your hard work!

jon-bown:
Hi @emeli-dral, thanks for the update! As far as I can see, this implementation matches the ItemMatch/NoMatch implementation. Is there anything in particular you can see that needs to change? In the unit tests I convert the lists to tuples before they are fed to the feature, same as the other PR so it won't work with lists inside the data frame. I will also resolve the merge conflicts in the meantime.

emeli-dral:
Hi @jon-bown,
You’re absolutely right, that was my mistake — thank you for pointing that out! I've merged main into your branch, and I'll proceed with the merge as soon as the tests pass.

Thanks again for your patience and your thorough work on this!

Issue #1315, opened 2024-09-23
Add new `JSONMatch()` descriptor to Evidently


elenasamuylova:
# Add a new `JSONMatch()` descriptor

About Hacktoberfest contributions: https://github.com/evidentlyai/evidently/wiki/Hacktoberfest-2024

**Description:**

The `JSONMatch()` descriptor compares a JSON object in a given column against a reference JSON object in the specified reference column and returns `True` or `False` based on whether the JSONs are identical.

**JSON Matching Criteria:**

- Both JSON objects must have the exact same set of keys.
- The values associated with each key must be identical.
- The order of keys does not matter.
- Whitespace outside of the actual values (e.g., spaces, tabs, or newlines) is ignored.
- If either of the inputs is an invalid JSON, they will not match, and the descriptor will return `False`.

**Reference:**
For a two-column descriptor implementation, use the `CustomPairColumnEval` template and check the `SemanticSimilarity` descriptor as an example (implement the `with_column` parameter the same way).

PR #1337, opened 2024-10-05
added JSONMatch Descriptor

ramachaitanya0:
This Change Adds JSONMatch Descriptor ( #1315) which compares two columns of a dataframe and checks whether the two objects in each row of the dataframe are matching JSON's or not.


src/evidently/descriptors/_registry.py
ramachaitanya0, line 1: I think I need to use FeatureDescriptor, I have used GeneralDescriptor. Will change it if the other changes are working fine and good to go for Merging.

emeli-dral: Hi @ramachaitanya0 , 
thank you for your contribution!

Your are right, FeatureDescriptor is a better much here. It will be much easier to resolve issues if you switch to  FeatureDescriptor. Please take a look at test results as well: [link to the error message](https://github.com/evidentlyai/evidently/actions/runs/11196989543/job/31172499476?pr=1337#step:9:505)

ramachaitanya0: Hi emeli, made the changes. New PR passes all the tests. What will be the next steps ?

src/evidently/descriptors/json_match_descriptor.py
emeli-dral, line 1: Hey @ramachaitanya0,

This looks great!

The only change I’d suggest is to remove the ```"display_name"``` class variable from the ```JSONMatch``` class. We want to allow users to receive a default display name if they don’t specify one. Instead, we prefer to generate the ```"display_name"``` from the corresponding feature.

To support this behavior, there’s one more change needed in the feature itself (I’ll add a comment there as well).

src/evidently/features/json_match_feature.py
emeli-dral, line 1: Hey @ramachaitanya0,

Thanks for your work! Here’s a list of changes to make so that users can get a default ```display_name``` if they don’t specify one:

1. Remove the ```display_name``` class variable:

We need the logic from the base class ```GeneratedFeatures``` method ```_create_column```, which handles selecting a default ```display_name``` when one is not provided.

2. Remove the ```name``` class variable:

To allow users to calculate multiple ```JSONMatch``` descriptors over the same DataFrame in a single report, we should avoid using a constant feature name. Keeping ```name``` as a class variable will result in errors. Instead, I suggest implementing a method like this:
```
def _feature_column_name(self):
    return f"JSON match for {self.first_column} and {self.second_column}"
```

3. Use the ```_feature_column_name``` method in the ```generate_feature``` method:

Update the code in the ```generate_feature``` method to use the new method for naming:
```
data[self._feature_column_name()] = data.apply(
    lambda x: compare_json_objects(x[self.first_column], x[self.second_column]), axis=1
)
return pd.DataFrame(data[self._feature_column_name()])
```

4. Update the ```_as_column``` method:

Modify the ```_as_column``` method to pass the generated column name and default display name as parameters:
```
def _as_column(self) -> "ColumnName":
        return self._create_column(
            self._feature_column_name(),
            default_display_name=f"JSON match for columns {self.first_column} and {self.second_column}",
        )
```

ramachaitanya0: Hi Emeli, thanks for the feedback. will make the changes

emeli-dral:
Hey @ramachaitanya0,
Thanks for adding the changes! Everything looks good to me. 

Could you please also add the new descriptor to the list of descriptors under the TextEvals on the [documentation page](https://docs.evidentlyai.com/reference/all-metrics#text-evals-1)? Once that’s done, we’ll be ready to proceed with the merge.

ramachaitanya0:
Hi Emeli, How can i edit the document ?

emeli-dral:
> Hi Emeli, How can i edit the document ?

@ramachaitanya0 the document is located in the same repository here: [all-metrics.md#text-evals](https://github.com/evidentlyai/evidently/blob/main/docs/book/reference/all-metrics.md#text-evals).

Feel free to use the other descriptors as a reference! :)

PR #1343, opened 2024-10-16
Update Mypy rules and version.

Liraim:


setup.py
```diff
@@ -84,7 +84,8 @@
             "setuptools==65.5.1; python_version < '3.12'",
             "setuptools==68.2.2; python_version >= '3.12'",
             "jupyter==1.0.0",
-            "mypy==0.981",
+            "mypy==1.1.1",
+            "pandas-stubs",
```
svetleo-evidently, line 88: version here

Liraim: added missing version.

PR #1374, opened 2024-11-25
Docs: change `TextOverviewPreset` to `TextEvals`

SkafteNicki:
On this page
https://docs.evidentlyai.com/presets/text-overview
there are references to a `TextOverviewPreset` class which does not seem to exist in the library. Instead I am fairly confident that instead it should be the `TextEvals` preset.


emeli-dral:
Hey @SkafteNicki,
Thanks a lot!

PR #1389, opened 2024-12-10
Metrics redesign

Liraim:


src/evidently/v2/datasets.py
```diff
@@ -0,0 +1,129 @@
...
+import pandas as pd
+
+from evidently import ColumnType
+from evidently.features.generated_features import GeneratedFeatures
+from evidently.options.base import Options
+
+
+@dataclasses.dataclass
+class ColumnInfo:
+    type: ColumnType
```
mike0sv, line 17: usually renamed as `type_` to not collide with `type`

src/evidently/v2/datasets.py
```diff
@@ -0,0 +1,129 @@
...
+class DatasetColumn:
+    type: ColumnType
+    data: pd.Series
+
+    def __init__(self, type: ColumnType, data: pd.Series) -> None:
+        self.type = type
+        self.data = data
+
+
+class Scorer:
```
mike0sv, line 40: need to think about naming, this class most definitely does not do "scoring" :)

src/evidently/v2/datasets.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __init__(self, columns: Dict[str, ColumnInfo]):
+        self._columns = columns
+
+    def get_column_type(self, column_name: str) -> ColumnType:
+        # TODO: implement
+        return ColumnType.Unknown
+
+
+class DatasetColumn:
```
mike0sv, line 31: dataclass?

src/evidently/v2/datasets.py
```diff
@@ -0,0 +1,129 @@
...
+        self.type = type
+        self.data = data
+
+
+class Scorer:
+    def __init__(self, alias: str):
+        self._alias = alias
+
+    @abc.abstractmethod
+    def generate_data(self, dataset: "Dataset") -> Union[DatasetColumn, Dict[str, DatasetColumn]]:
```
mike0sv, line 45: I try to avoid unions, because every it means i will need to do instance check somewhere. How about we use general interface with a dict, but have something like `SingleColumnScorer` that will delegate this method to another abs method that returns a single  `DatasetColumn`?

src/evidently/v2/datasets.py
```diff
@@ -0,0 +1,129 @@
...
+        raise NotImplementedError()
+
+    @property
+    def alias(self) -> str:
```
mike0sv, line 49: why this is not a field?

src/evidently/v2/metrics/base.py
```diff
@@ -0,0 +1,200 @@
...
+from evidently.utils.dashboard import TemplateParams
+from evidently.utils.dashboard import inline_iframe_html_template
+from evidently.v2.datasets import Dataset
+from evidently.v2.datasets import DatasetColumn
+
+
+class MetricResult:
+    _widget: Optional[List[BaseWidgetInfo]] = None
+
+    def set_widget(self, widget: List[BaseWidgetInfo]):
```
mike0sv, line 32: seems redundant

Issue #1410, opened 2025-01-08
Update evidently hashlib usage for FIPS-Compliant Systems and Security Best Practices


DueViktor:
From python 3.9 `hashlib` introduced the `usedforsecurity` argument:
> Changed in version 3.9: All hashlib constructors take a keyword-only argument usedforsecurity with default value True. A false value allows the use of insecure and blocked hashing algorithms in restricted environments. False indicates that the hashing algorithm is not used in a security context, e.g. as a non-cryptographic one-way compression function.

`evidently` use hashing where the purpose is indeed _not for security_ purposes. This should be specifed in the code. My guess is that pretty much all federal systems in the world would have this issue.
> Federal Information Processing Standards (FIPS) 140-2 is a mandatory standard for the protection of sensitive or valuable data within Federal systems. - https://www.wolfssl.com/license/fips/

Without this fix organisations following certain security standard (FIPS compliance could be an example) are not able to use this package. 

The same fix has been applied to a range of key repositories:
- https://github.com/streamlit/streamlit/issues/7120
- https://github.com/mlflow/mlflow/issues/10106
- https://github.com/huggingface/transformers/issues/27034
- https://github.com/chroma-core/chroma/issues/1672

# Proposed solution

- Explicitly set `usedforsecurity=False` when calling `hashlib` constructors where the hashing is **not security-related**.
- Since `usedforsecurity` is only available from Python 3.9 onwards and `evidently` supports Python 3.8+, implement logic to handle this difference:
  - For Python 3.8: Call `hashlib` without `usedforsecurity`.  
  - For Python 3.9+: Specify `usedforsecurity=False`.

I will submit a PR implementing this change. Let me know if there are other concerns or suggestions for handling this!

jbbqqf:
Hi! Triaging older issues this week — I think this one has shipped and can be closed.

`hashlib.md5` calls in evidently already pass `usedforsecurity=False`. The keyword is centralized at [`src/evidently/pydantic_utils.py:43`](https://github.com/evidentlyai/evidently/blob/main/src/evidently/pydantic_utils.py#L43):

```python
md5_kwargs = {"usedforsecurity": False}
```

[...]

PR #1411, opened 2025-01-08
FIPS compliant

DueViktor:
Provides a fix for problem raised in https://github.com/evidentlyai/evidently/issues/1410


DueViktor:
I'll fix the linter later today.

PR #1533, opened 2025-03-31
Remove scipy deprecation warning

tomknip:
In our organisation we want to start using Evidently. However, we have strict tests on Deprecation warnings not being allowed. Currently, there is a Deprecation warning within the Evidently package around Scipy. This PR solves that Deprecation warning. Discussed in this [issue](https://github.com/evidentlyai/evidently/issues/1534)


emeli-dral:
Hi @tomknip ,
Thanks a lot!

Issue #1627, opened 2025-05-30
MAE tests are missing in report


emacha:
# Overview

Some of the regression metrics (`MAE` and `MeanError` at least) are not being executed when running a report.

# Example code

```python
import pandas as pd
from evidently.future.datasets import Dataset, DataDefinition, Regression
from evidently.future.metrics import MAE, RMSE
from evidently.future.report import Report
from evidently.future import tests

data = pd.DataFrame({
    "id_col": [1, 2, 3],
    "actual": [10.0, 20.0, 30.0],
    "predicted": [11.0, 19.0, 29.0]
})
regression_ = [Regression(target="actual", prediction="predicted")]
data_definition = DataDefinition(id_column="id_col", regression=regression_)
dataset = Dataset.from_pandas(data, data_definition=data_definition)
report = Report(
    [
        # 3 tests for RMSE, 1 test for MAE -> Expected: 4 tests in total
        RMSE(tests=[tests.lt(100), tests.gt(0), tests.not_eq(0)]),
        MAE(tests=[tests.lt(100)]),
    ],
)
snapshot = report.run(dataset).dict()
assert len(snapshot["tests"]) == 4, "Expected 4 tests, but got {}".format(len(snapshot["tests"]))

>>> assert len(snapshot["tests"]) == 4, "Expected 4 tests, but got {}".format(len(snapshot["tests"]))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>>> AssertionError: Expected 4 tests, but got 3
```

# Versions

Evidently: 0.7.6
Python: 3.12.7

emacha:
@elenasamuylova Yeah that worked, thanks!

Do you think you could raise an exception in this case? It's a pretty large footgun.

bharath03-a:
Hi @elenasamuylova!
I’ve created a PR that raises an explicit exception when tests= is passed to MAE instead of mean_tests or std_tests. Happy to adjust for any suggested changes. Let me know your thoughts!

PR #1661, opened 2025-07-16
Raising Exception for Missing `mean_tests` or `std_tests` in MAE Metric

bharath03-a:
### Description

This PR addresses issue #1627. Previously, passing `tests=[...]` to `MAE()` silently ignored the test, leading to unexpected results in reports. This could confuse users, especially when writing assertions based on the number of tests executed.

### Fix

- Added a check in the `MAE` class constructor to raise a `ValueError` if `tests` is provided instead of `mean_tests` or `std_tests`.
- The exception clearly instructs the user to use `mean_tests` or `std_tests` depending on what they want to test.

I had considered using a pydantic.Config class with extra = "forbid" to prevent accidental usage. It felt like a cleaner and more generalized solution for validating input.

However, I wasn't entirely sure if this would break other internal logic or dynamic field handling elsewhere in the codebase.

If this kind of config-based validation is acceptable or preferred, I'm happy to explore replacing the current manual ValueError check with a more declarative Pydantic approach. Let me know what you think!

### DIFF
--- a/src/evidently/metrics/regression.py
+++ b/src/evidently/metrics/regression.py
@@ -136,6 +136,11 @@ class MAE(MeanStdMetric):
     error_distr: bool = True
     error_normality: bool = False
 
+    def __init__(self, **kwargs):
+        if "tests" in kwargs:
+            raise ValueError("'tests' is not a valid argument for MAE. Did you mean 'mean_tests=' or 'std_tests='?")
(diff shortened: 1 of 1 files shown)


bharath03-a:
@mike0sv yeah, I agree too adding pydantic config makes sense. Should I go ahead try that out, since this PR is already merged?

PR #1684, opened 2025-08-04
feat(docker): Allow installing optional extras in service image in order to support S3 and GCS usage

isreolv:
**What does this PR do?**
This PR introduces an INSTALL_EXTRAS optional build argument to the Dockerfile.service. This allows users to build the service image with optional dependencies, such as s3fs for S3 support or openai for LLM evaluations. It also adds an example to the documentation on how to use Evidently AI Docker image with S3/GCS.

**Why is this change needed?**
Currently, the default Docker image only contains core dependencies. Users who want to connect the Evidently UI workspace to a data storage service on S3 or GCS will encounter an error because libraries like s3fs are not installed. This change provides a clean, documented way to build a properly configured image without forcing all optional dependencies on every user.

How to use this feature:
To build the image with support for S3 and GCS, a user can now run:

```bash
docker build
  --build-arg INSTALL_EXTRAS="[fsspec]" \
  -f docker/Dockerfile.service \
  -t evidently/evidently-service:custom .
```

It's also possible to use it to install any optional dependency, such as the dependencies required for LLM evaluations:

```bash
docker build \
  --build-arg INSTALL_EXTRAS="[llm]" \
  -f docker/Dockerfile.service \
  -t evidently/evidently-service:llm .
```

This resolves the issue of needing to manually edit the Dockerfile for common use cases. I have also included a documentation snippet to explain this new functionality to users.

### DIFF
--- a/examples/service/README.md
+++ b/examples/service/README.md
@@ -8,7 +8,8 @@ This example demonstrates how to run an **Evidently UI service** and connect to 
 
 - `run_service.sh` — Script to start the Evidently UI service in Docker.
 - `remote_demo_project.py` — Python script that uploads a demo bike rentals monitoring project to the running Evidently service.
-- `workspace_tutorial.ipynb` — Jupyter notebook with Evidently UI API tutorial
+- `workspace_tutorial.ipynb` — Jupyter notebook with Evidently UI API tutorial.
+- `docker_s3_tutorial.ipynb` — A complete tutorial on how to use the Evidently UI official docker image with S3 and GCS.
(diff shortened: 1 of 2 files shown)


emeli-dral:
Hi @israolive ,
Thanks for the thoughtful implementation! This is a really strong improvement.
Introducing INSTALL_EXTRAS as a build arg is flexible for advanced users, but doesn’t complicate the base case or increase image size unnecessarily. That’s exactly how optional dependencies should be handled.

Also appreciate that the tutorial is already in place. 
Great work, merging this in.

PR #1732, opened 2025-10-26
Allow Plotly v6 in dependencies

tomknip:
Increase the scope of the Plotly dependency from v5 to also being able to update to v6.


tomknip:
@DimaAmega Indeed, was expecting such a thing to happen already. Did not have the time to look into it however..

Issue #1706, opened 2025-08-28
GroupBy(ValueDrift) display_name not shown properly


shpp9212:
Hi,

When I use

```
report = Report([
    GroupBy(MaxValue(column="column1"), "column2"),
    DataDriftPreset(),
    RegressionPreset()
],
include_tests=True)
```
the resulting report displays properly, "Maximum value of 'column1' group by 'column2' for label: 'abc'"

On the other hand, when I use 

```
report = Report([
    GroupBy(ValueDrift(column="column1"), "column2"),
    DataDriftPreset(),
    RegressionPreset()
],
include_tests=True)
```

the resulting report displays just "Drift in column 'column1'".

Could you fix this issue? Thank you for developing such a great framework.

PR #1785, opened 2025-12-26
Fix Groupby display name for Valuedrift metrics

Sudharsan-T:
### Problem

`GroupBy(ValueDrift(...))` does not include the group-by column name and label in the report display name, while other metrics such as `MaxValue` do.

This results in inconsistent output:

* `GroupBy(MaxValue(...))` → includes group-by column and label
* `GroupBy(ValueDrift(...))` → shows only `Drift in column '<column>'`

Reported in #1706.

---

### Root cause

`GroupByMetricCalculation` relied on metric-specific internal display name generation. For `ValueDrift`, this display name does not carry group-by context, so the column name and label were lost.

---

### Fix

Build the display name explicitly in `GroupByMetricCalculation.display_name()` using:

* the base display name of the underlying metric
* the group-by column name
* the group label

This makes the behavior consistent across all `GroupBy` metrics.

---

### Implementation

**File:**

```
src/evidently/metrics/group_by.py
```

The display name construction was moved to the GroupBy calculation layer instead of relying on metric-specific behavior.

---

### Result

`GroupBy(ValueDrift(...))` now displays:

```
Drift in column '<column>' group by '<group_column>' for label: '<label>'
```

---

### Testing

* Reproduced the issue locally using the example from #1706
* Verified output for both `MaxValue` and `ValueDrift`
* Formatting and lint checks pass

### DIFF
--- a/src/evidently/metrics/group_by.py
+++ b/src/evidently/metrics/group_by.py
@@ -35,8 +35,6 @@ class GroupByMetricCalculation(MetricCalculation[TResult, GroupByMetric]):
     def calculate(self, context: "Context", current_data: Dataset, reference_data: Optional[Dataset]):
         curr = current_data.subdataset(self.metric.column_name, self.metric.label)
         ref = reference_data.subdataset(self.metric.column_name, self.metric.label) if reference_data else None
-        dn = self.calculation.display_name
-        self.calculation.display_name = _patched_display_name(dn, self.metric)  # type: ignore[method-assign]
         res = self.calculation.calculate(context, curr, ref)
         if isinstance(res, tuple):
             curr_res, ref_res = res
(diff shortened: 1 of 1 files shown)


DimaAmega:
Hi @Sudharsan-T 
image_2026-01-05_15-12-20 (2)

Here we see less information than before, so this also needs to be fixed

Check [html-visual-testing-playwright-report](https://github.com/evidentlyai/evidently/actions/runs/20521268655/artifacts/5024868557)

Liraim:
Hi @Sudharsan-T,

Thanks for the PR and for working on this improvement. 

As @DimaAmega mentioned, there is one important issue I want to highlight, and why it occurs:
- GroupBy wrapper relies on renders of underlying metrics to create a widget in the report. To customize the display name, it replaces the `display_name()` with a patched version (using `_patched_display_name`)
- In this PR, this patched method was removed, which means every metric rendered through GroupBy loses the patched `display_name` and therefore its group-by-specific label.

To fix this properly, I think the following changes are needed:
- bring back the use  of `_patched_display_name` in the group-by wrapper
- update the render method in `ValueDriftCalculation` (`src/evidently/metrics/column_statistics.py:588`) to use `self.display_name()` method instead of the hardcoded string
- update `self.display_name()` to return proper string (it seems now it uses a different text from what is used in render).

This should restore the expected behavior for all metrics, and ValueDrift metrics correctly display the group-by text when used inside GroupBy.

PR #1820, opened 2026-02-07
Fail early on empty input data in Report.run

Si-ra-kri:
### Summary
Adds early validation to fail with a clear error when empty DataFrames
are passed to Report.run.

### Motivation
Passing empty input data currently results in a ZeroDivisionError deep
inside legacy drift calculations, which is confusing for users.

### Changes
- Added input validation for empty DataFrames
- Added a unit test covering this case

### Scope
- No API changes
- No metric logic changes

### DIFF
--- a/tests/report/test_report_input_validation.py
+++ b/tests/report/test_report_input_validation.py
@@ -2,11 +2,10 @@ import pandas as pd
 import pytest
 
 from evidently import Report
-from evidently.metrics import ColumnSummaryMetric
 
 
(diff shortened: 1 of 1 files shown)


Si-ra-kri:
Added a minimal input-validation test that avoids metric imports so it works in the CI environment.
All local tests and pre-commit checks pass, updating the PR for re-run.

DimaAmega (approved):

PR #1841, opened 2026-03-01
Support target_names in classification metrics renderers

### LINKED ISSUE #578: Support 'target_names' in Classification Metrics
If ```target_names``` is specified in the ```column_mapping``` it needs to be used in the following metrics:
- ```ClassificationQualityByClass()```
- ```ClassificationClassSeparationPlot()```
- ```ClassificationProbDistribution()```
- ```ClassificationQualityByFeatureTable()```

Screenshots below illustrate where ```target_names``` are supposed to appear on plots:

1. ClassificationQualityByClass()
ClassificationQualityByClass

2. ClassificationClassSeparationPlot()

ClassificationClassSeparationPlot

3. ClassificationProbDistribution()

ClassificationProbDistribution

4. ClassificationQualityByFeatureTable()
ClassificationQualityByFeatureTable

Br1an67:
Fixes #578

When `target_names` is specified in `column_mapping` as a dict mapping raw class values to human-readable names, three classification metric renderers were ignoring it and displaying raw class values. This PR adds `target_names` support to:

- **ClassificationClassSeparationPlot** — tab names now use display names
- **ClassificationProbDistribution** — distribution plot titles and legend labels now use display names
- **ClassificationQualityByFeatureTable** — tab titles and scatter plot trace names now use display names

The existing `ClassificationQualityByClass` metric already supported `target_names` — the same pattern is applied here.

**Changes:**
- Added `target_names: Optional[TargetNames]` field to each result class (defaults to `None` for backward compatibility)
- Passed `dataset_columns.target_names` through the calculate → result → renderer pipeline
- Added a `_resolve_target_name()` helper that maps raw labels to display names when a dict mapping is provided

### DIFF
--- a/src/evidently/legacy/metrics/classification_performance/class_separation_metric.py
+++ b/src/evidently/legacy/metrics/classification_performance/class_separation_metric.py
@@ -18,6 +18,7 @@ from evidently.legacy.metric_results import df_from_column_scatter
 from evidently.legacy.metric_results import raw_agg_properties
 from evidently.legacy.model.widget import BaseWidgetInfo
 from evidently.legacy.options.base import AnyOptions
+from evidently.legacy.pipeline.column_mapping import TargetNames
 from evidently.legacy.renderers.base_renderer import MetricRenderer
 from evidently.legacy.renderers.base_renderer import default_renderer
 from evidently.legacy.renderers.html_widgets import TabData
@@ -40,6 +41,7 @@ class ClassificationClassSeparationPlotResults(MetricResult):
         }
 
     target_name: str
+    target_names: Optional[TargetNames] = None
 
     current: Optional[ColumnScatterOrAgg] = None
     current_raw, current_agg = raw_agg_properties("current", ColumnScatter, ColumnAggScatter, True)
@@ -107,6 +109,7 @@ class ClassificationClassSeparationPlot(UsesRawDataMixin, Metric[ClassificationC
                 current=column_scatter_from_df(current_plot, True),
                 reference=column_scatter_from_df(reference_plot, True),
                 target_name=target_name,
+                target_names=dataset_columns.target_names,
             )
         current_plot = prepare_box_data(current_plot, target_name, prediction_names.tolist())
         if reference_plot is not None:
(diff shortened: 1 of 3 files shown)


src/evidently/legacy/metrics/classification_performance/class_separation_metric.py
```diff
@@ -115,9 +118,18 @@ def calculate(self, data: InputData) -> ClassificationClassSeparationPlotResults
             current=current_plot,
             reference=reference_plot,
             target_name=target_name,
+            target_names=dataset_columns.target_names,
         )
 
 
+def _resolve_target_name(label, target_names: Optional[TargetNames]) -> str:
+    if target_names is not None and isinstance(target_names, dict):
+        resolved = target_names.get(label) or target_names.get(int(label)) if isinstance(label, str) else target_names.get(label)  # type: ignore[arg-type]
```
Nik-Reddy, line 127: Two things here. First, this only triggers when target_names is a dict, but TargetNames is typed as Union[List, Dict]. If someone passes a list, the resolver does nothing and falls through to str(label). Not sure if list-based target_names is actually used in practice, but worth checking.

Second, the or on this line can swallow falsy mapped values. If target_names maps a label to 0 or an empty string, the or skips it and tries int(label) instead. And if label is a non-numeric string like "cat", int(label) raises ValueError. Safer to check `is None` explicitly rather than relying on truthiness.

src/evidently/legacy/metrics/classification_performance/class_separation_metric.py
```diff
@@ -150,7 +162,7 @@ def render_html(self, obj: ClassificationClassSeparationPlot) -> List[BaseWidget
                 target_name,
                 color_options=self.color_options,
             )
-        tabs = [TabData(name, widget) for name, widget in tab_data]
+        tabs = [TabData(_resolve_target_name(name, metric_result.target_names), widget) for name, widget in tab_data]
```
Nik-Reddy, line 165: Tab names are resolved here which is good, but the inner traces built by class_separation_traces_raw and class_separation_traces_agg still use str(label) for name and legendgroup. So the tab heading will show the display name but the plot legend inside the tab still shows the raw class value. Might want to pass target_names into those trace-building helpers too.

src/evidently/legacy/metrics/classification_performance/quality_by_feature_table.py
```diff
@@ -173,18 +175,28 @@ def calculate(self, data: InputData) -> ClassificationQualityByFeatureTableResul
...
             target_name=target_name,
+            target_names=dataset_columns.target_names,
         )
 
 
 @default_renderer(wrap_type=ClassificationQualityByFeatureTable)
 class ClassificationQualityByFeatureTableRenderer(MetricRenderer):
+    @staticmethod
+    def _resolve_target_name(label, target_names: Optional[TargetNames]) -> str:
+        if target_names is not None and isinstance(target_names, dict):
```
Nik-Reddy, line 186: Same resolver function as the other two files. If the operator precedence or dict-only issue needs fixing, it has to be done in three places. Pulling this into something like evidently.legacy.utils or a shared base would keep it maintainable.

PR #1849, opened 2026-03-26
Pin evidently versions in Grafana examples.

Liraim:


Liraim:
Changes already merged in another PR.

PR #1852, opened 2026-03-27
Fix `list_projects()` method when using a workspace with an S3 path

### LINKED ISSUE #1848: Projects list is always empty when initializing an S3 workspace
There seems to be a bug in the `Workspace.list_projects()` method when the workspace is configured with an S3 URI. The method returns an empty list, even when the workspace contains multiple projects.

```python
ws = Workspace(path="s3://bucket/evidently")
print(ws.list_projects())
# Output: []
# Expected a list of projects
```

The root cause seems to lie in the code fragment below, where a dot `.` is passed to list the directory contents. While this works for POSIX file systems, it does not behave the same in S3. In S3, `.` is treated as part of the object key, meaning `/foo` and `/foo/.` are seen as distinct prefixes.

https://github.com/evidentlyai/evidently/blob/67d15117bcfefb81f2b1bf558cc40ee779ea3766/src/evidently/ui/storage/local/base.py#L73

To confirm this behavior, I modified [fslocation.py](https://github.com/evidentlyai/evidently/blob/67d15117bcfefb81f2b1bf558cc40ee779ea3766/src/evidently/ui/service/storage/fslocation.py) to log the path. The example below shows that, indeed, a `.` is being appended to the S3 path..

```python
def listdir(self, path: str):
    try:
        fullpath = posixpath.join(self.path, path)
        print(fullpath)

# Outputs: bucket/artifacts/evidently/.
# Expected bucket/artifacts/evidently/
```

In another part of the codebase, directory listing is handled correct by passing an empty string instead of a `.`.

https://github.com/evidentlyai/evidently/blob/67d15117bcfefb81f2b1bf558cc40ee779ea3766/src/evidently/ui/service/storage/local/base.py#L124

armandduijn:
Fix `list_projects()` method when using a workspace with an S3 path. S3 doesn't doesn't interpret `.` as "current directory", but as a different key.

Fixes https://github.com/evidentlyai/evidently/issues/1848

### DIFF
--- a/src/evidently/ui/storage/local/base.py
+++ b/src/evidently/ui/storage/local/base.py
@@ -70,7 +70,7 @@ class LocalState:
 
     def list_projects(self) -> List[ProjectID]:
         projects = []
-        for p in self.location.listdir("."):
+        for p in self.location.listdir(""):
             if not UUID_REGEX.match(p):
                 continue
             projects.append(ProjectID(p))


Liraim (approved):
Thank. you for your contribution in the project!
Changes look good.

PR #1862, opened 2026-04-21
fix: replace hardcoded fill value with dynamic min_nonzero/10 in get_binned_data

### LINKED ISSUE #334: The fixed value for feel_zeroes in get_binned_data may lead to deviation in some case.

mukund1985:
## Summary

Fixes #334

The zero-fill logic in `get_binned_data` used a hardcoded threshold and fallback of `0.0001`. When all non-zero percentages in a distribution are smaller than `0.0001` (common with large datasets or rare categories), the fill value becomes **larger** than legitimate data values. This makes KL-divergence and other stattest calculations incorrect — the fill is supposed to be a negligible epsilon, not a dominant value.

**Root cause:**
```python
# BEFORE — fill can exceed real values when min_nonzero <= 0.0001
np.place(reference_percents, reference_percents == 0,
    min(reference_percents[reference_percents != 0]) / 10**6
    if min(reference_percents[reference_percents != 0]) <= 0.0001
    else 0.0001)
```

**Fix:**
```python
# AFTER — always proportional to the smallest real value in that array
ref_nonzero = reference_percents[reference_percents != 0]
if len(ref_nonzero) > 0:
    np.place(reference_percents, reference_percents == 0, min(ref_nonzero) / 10)
```

`min_nonzero / 10` is guaranteed strictly smaller than any real data value at any scale. The empty-array guard prevents errors when one side has no non-zero entries.

## Changes

- `src/evidently/legacy/calculations/stattests/utils.py` — pandas implementation
- `src/evidently/legacy/spark/calculations/stattests/utils.py` — Spark implementation

Both use identical logic with their respective parameter names (`feel_zeroes` / `fill_zeroes`).

## Test plan

- [x] All 156 existing stattest unit tests pass (`pytest tests/multitest/metrics/test_data_drift.py tests/stattests/ -v`)
- [x] Smoke test confirms fill value equals exactly `min_nonzero / 10` and old hardcoded `0.0001` is gone
- [x] Verified fix handles edge case where all values in one array are non-zero (empty guard)

### DIFF
--- a/src/evidently/legacy/calculations/stattests/utils.py
+++ b/src/evidently/legacy/calculations/stattests/utils.py
@@ -42,20 +42,12 @@ def get_binned_data(
         current_percents = np.array([current_feature_dict[key] / len(current_data) for key in keys])
 
     if feel_zeroes:
-        np.place(
-            reference_percents,
-            reference_percents == 0,
-            min(reference_percents[reference_percents != 0]) / 10**6
-            if min(reference_percents[reference_percents != 0]) <= 0.0001
-            else 0.0001,
-        )
-        np.place(
-            current_percents,
-            current_percents == 0,
-            min(current_percents[current_percents != 0]) / 10**6
-            if min(current_percents[current_percents != 0]) <= 0.0001
-            else 0.0001,
-        )
+        ref_nonzero = reference_percents[reference_percents != 0]
+        cur_nonzero = current_percents[current_percents != 0]
+        if len(ref_nonzero) > 0:
+            np.place(reference_percents, reference_percents == 0, min(ref_nonzero) / 10)
+        if len(cur_nonzero) > 0:
+            np.place(current_percents, current_percents == 0, min(cur_nonzero) / 10)
 
     return reference_percents, current_percents
 
--- a/src/evidently/legacy/spark/calculations/stattests/utils.py
+++ b/src/evidently/legacy/spark/calculations/stattests/utils.py
@@ -52,19 +52,11 @@ def get_binned_data(
         current_percents = current_percents / current_percents.sum()
 
     if fill_zeroes:
-        np.place(
-            reference_percents,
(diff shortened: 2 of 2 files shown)


mukund1985:
Hey, just flagging — ran the existing test suite locally and everything passes. The fix is pretty minimal, just handling that edge case. Happy to change the approach if there's a better way to do it, just let me know.

mukund1985:
@Liraim — would appreciate a review when you get a chance. Tests pass locally, happy to make any changes needed.

mukund1985:
@DimaAmega — looks like CI hasn't triggered yet, could you approve the workflow run when you get a chance?

Issue #1854, opened 2026-04-02
Harden GitHub Actions workflows against CI/CD supply chain attacks


strickvl:
## Context

A recent automated attack campaign ([hackerbot-claw](https://www.stepsecurity.io/blog/hackerbot-claw-github-actions-exploitation)) targeted CI/CD pipelines across major open source repositories (Microsoft, DataDog, CNCF projects), achieving remote code execution in multiple targets and exfiltrating a `GITHUB_TOKEN` with write permissions from one repository with 140k+ stars.

I audited Evidently's GitHub Actions workflows against the specific attack vectors used in this campaign. **No signs of compromise were found**, but several hardening opportunities exist.

I'd be happy to open a PR for these fixes, but since some changes (particularly adopting Zizmor) would affect your CI/CD process, I thought it best to open an issue for discussion first.

### 1. Script Injection via `${{ }}` in `run:` blocks (Medium-High)

Several workflows interpolate user-controlled values directly into shell commands, which enables [script injection](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections). An attacker can craft a branch name containing shell metacharacters to execute arbitrary commands.

**Affected locations:**

| File | Lines | Expression | Notes |
|------|-------|-----------|-------|
| `main.yml` | 172, 384, 441 | `${{ github.event.pull_request.head.ref }}` in shell | Branch name → shell injection |
| `main.yml` | 274 | `${{ github.event.pull_request.head.ref }}` as CLI arg | Branch name → argument injection |
| `main.yml` | 78, 84, 90, 96-97 | `${{ steps.changed-files.outputs.*_files }}` in echo | Filenames from PR → shell injection |
| `deploy-artifacts-to-github-pages.yml` | 70 | `${{ github.event.workflow_run.head_branch }}` in shell | **Most concerning** — `workflow_run` runs with base repo's full write permissions and secrets |
| `docker.yml` | 16 | `echo ${{ secrets.DOCKER_PWD }}` | Secret expanded inline in shell |

**Fix:** Assign `${{ }}` values to environment variables, then reference them as `"$VAR"` in shell. Environment variables are treated as data, not code:

```yaml
# Before (vulnerable):
run: |
  BRANCH_NAME="${{ github.event.pull_request.head.ref }}"

# After (safe):
env:
  BRANCH_NAME: ${{ github.event.pull_request.head.ref }}
run: |
[...]

PR #1864, opened 2026-04-21
Harden workflows against script injection (#1854)

strickvl:
Addresses the script-injection findings from #1854 (section 1 of the audit). Scoped exactly as discussed in [this comment](https://github.com/evidentlyai/evidently/issues/1854#issuecomment-4291997741) — action pinning will follow in a separate PR. `permissions:` hardening, Zizmor adoption, and the Docker secret handling are left for your internal review as requested.

## Summary

Every `${{ … }}` expression that previously landed directly inside a `run:` block has been moved to a step-level `env:` variable (or, for the composite action, into a structured input). Shell sees environment variables as data — metacharacters inside them can't re-enter the parser.

## Files changed

| File | Sites | Value formerly in shell |
|---|---|---|
| `main.yml` | 6 | `github.event.pull_request.head.ref` (×3), `steps.changed-files.outputs.*_all_changed_and_modified_files` (×5 expressions across 4 steps), `github.event.pull_request.head.repo.full_name`, `env.ARTIFACT_PREFIX` (now used as plain `$ARTIFACT_PREFIX` since it's a workflow-level env var) |
| `deploy-artifacts-to-github-pages.yml` | 1 | `github.event.workflow_run.head_branch`, `github.repository` |
| `share-actions/generate-api-reference/action.yml` | 1 (refactor) | See below |
| `release.yml` | 2 | Updated call site for composite action; also moved `github.ref_name` out of the version-check shell (see note) |

### Composite action refactor

`generate-api-reference/action.yml` previously accepted a single free-form `flags` string and expanded it with `${{ inputs.flags }}` inside a `run:` block — which meant any unsafe value in the caller's flags string (e.g., a branch name) hit the shell as tokens, not data. Refactored to four typed inputs (`local-source-code`, `github-repo`, `git-revision`, `output-prefix`), with argv built via a bash array so values are passed as single quoted arguments. Both callers (`main.yml` and `release.yml`) updated.

### One extra fix beyond the original audit

While working on `release.yml` I noticed that the version-check step (`main.yml` → sorry, `release.yml` lines 20-28) interpolated `github.ref_name` directly into the shell. Git tag names can legally contain `"` among other characters, so a tag like `v1.0"; rm -rf /; echo "` could escape the quotes and execute. Same class of issue as the rest of section 1, so I included it here — happy to split out if you'd rather keep this PR strictly to the audit list.

## What's explicitly left for later

- **Pinning actions to stable versions** — separate PR, per the comment thread.
- **`permissions: read-all`** — not touched.
- **`tj-actions/changed-files@v42`** — not touched in this PR (pinning PR will address).
- **`docker.yml` secret handling** — deferred to internal review as requested.
- **Zizmor adoption** — deferred to internal discussion as requested.

## Test plan

- [ ] CI on this PR should exercise `main.yml` end-to-end (the changed-files job, linter, tests, UI builds, API reference generation, Playwright runs).
- [ ] `release.yml` is tag-triggered so can't be directly exercised from a PR — logic-preserving refactor only (env-var rename; no control-flow change).
- [ ] `deploy-artifacts-to-github-pages.yml` runs on `workflow_run` after CI completes; behavior preserved (env-var move only).

### DIFF
--- a/.github/share-actions/generate-api-reference/action.yml
+++ b/.github/share-actions/generate-api-reference/action.yml
@@ -1,8 +1,21 @@
 name: Generate docs and upload artifact
 inputs:
-  flags:
-    description: 'Flags to pass to the generate.py script'
-    required: true
+  local-source-code:
+    description: 'Whether to generate docs from the locally checked out source code'
+    required: false
+    default: 'false'
+  github-repo:
+    description: 'GitHub repository (owner/name) used for source links in generated docs'
+    required: false
+    default: ''
+  git-revision:
+    description: 'Git revision (branch or tag) used for source links in generated docs'
+    required: false
+    default: ''
+  output-prefix:
+    description: 'Output prefix for generated artifacts'
+    required: false
+    default: ''
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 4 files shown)


Liraim (approved):
Thanks for your contribution!
The changes look good.

PR #1880, opened 2026-05-14
Fix/1616 pd na in by label count

### LINKED ISSUE #1616: Preset Failures with pandas NA class pd.NA (pydantic ValidationError)
## Issue

The presets do not work with [the pandas native missing value class](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.NaT.html) `pd.NA` (alias for `<NA>`), and will throw a `pydantic ValidationError`

Example:

```python
import pandas as pd

from evidently import DataDefinition, Dataset
from evidently.presets import DataDriftPreset, DataSummaryPreset

df = pd.DataFrame(
    {"a": pd.Series(["x", "y", "z", pd.NA], dtype="str")}
)

definition = DataDefinition(categorical_columns=['a'])
report = Report([DataSummaryPreset(),])
data =  Dataset.from_pandas(df, definition)

evaluation = report.run(current_data=data)
```

This will fail with:

```log

│ /Users/xxxxx/lib/python3.12/site-packages/evidently/core/metric_types.py:1004  │
│ in result                                                                                        │
│                                                                                                  │
│   1001 │   │   raise NotImplementedError                                                         │
│   1002 │                                                                                         │
│   1003 │   def result(self, count: Dict[Label, Value], shares: Dict[Label, Value]) -> ByLabelCo  │
│ ❱ 1004 │   │   return ByLabelCountValue(                                                         │
│   1005 │   │   │   counts={                                                                      │
│   1006 │   │   │   │   k: SingleValue(                                                           │
│   1007 │   │   │   │   │   value=v,                                                              
| /Users/xxxxx/lib/python3.12/site-packages/pydantic/v1/main.py:347 in __init__  │
│                                                                                                  │
│    344 │   │   # Uses something other than `self` the first arg to allow "self" as a settable a  │
│    345 │   │   values, fields_set, validation_error = validate_model(__pydantic_self__.__class_  │
│    346 │   │   if validation_error:                                                              │
│ ❱  347 │   │   │   raise validation_error                                                        │
│    348 │   │   try:                                                                              │
│    349 │   │   │   object_setattr(__pydantic_self__, '__dict__', values)                         │
│    350 │   │   except TypeError as e:  

ValidationError: 4 validation errors for ByLabelCountValue
counts -> __key__
  value is not a valid integer (type=type_error.integer)
counts -> __key__
  str type expected (type=type_error.str)
shares -> __key__
  value is not a valid integer (type=type_error.integer)
shares -> __key__
  str type expected (type=type_error.str)
```

## Motivation

The workaround would be to convert the `pd.NA`to `None`. 

However, ideally I would like to run the  exploration before performing any data processing at all, on the raw data. This issue prevents the initial data monitoring from running.

## Additional Info

Tested with `DataSummaryPreset` and `DataDriftPreset`

Evidently version: `0.7.4`
Pandas version: `2.1.4`
Python Version: `3.12.9`

ugbotueferhire:
Fixes #1616.

## Description
`DataSummaryPreset` and `DataDriftPreset` crashed with a `pydantic.ValidationError` whenever a categorical column contained `pd.NA` — pandas' native missing-value sentinel for nullable dtypes (`string`, `Int64`, `boolean`, etc.). Users had to convert `pd.NA` to `None` before passing data in, defeating the point of running exploratory metrics on raw data.

This PR makes both presets handle `pd.NA` transparently. Missing values are treated as missing — counted by `MissingValueCount`, never surfaced as a phantom `<NA>` label in `ByLabelCountValue`. The fix is three small, layered changes (one direct cause + two defensive guards so the same class of bug can't quietly resurface in future `ByLabelCount` metrics). No public API changes; no `Label` type change.
```python
import pandas as pd
from evidently import DataDefinition, Dataset, Report
from evidently.presets import DataSummaryPreset

df = pd.DataFrame({"a": pd.Series(["x", "y", "z", pd.NA], dtype="string")})
ds = Dataset.from_pandas(df, DataDefinition(categorical_columns=["a"]))
Report([DataSummaryPreset()]).run(current_data=ds)
# Before: pydantic.ValidationError on ByLabelCountValue (counts/shares __key__)
# After:  runs cleanly; pd.NA is treated as missing, not as a label.
```

The same failure hit `DataDriftPreset`. Both presets now pass.

## Root cause

Three cooperating defects:

1. **`UniqueValueCountCalculation._all_unique_values`** used `pd.Series.unique()`, which preserves `pd.NA` / `np.nan`. The NA sentinel was then carried into the result dict as a label key.
2. **`ByLabelCountValue`** lacked the `@validator(..., pre=True)` key-coercion that its sibling `ByLabelValue` has, so non-`Label`-typed dict keys (e.g. `np.int64`, `pd.NA`) reached pydantic uncoerced.
3. **`convert_types`** relied on `np.isnan(val)`, which raises `TypeError: boolean value of NA is ambiguous` on `pd.NA`.

## Changes

- **`src/evidently/metrics/column_statistics.py`** — `UniqueValueCountCalculation._all_unique_values` now calls `.dropna().unique()`, aligning the unique-value set with the existing `value_counts(dropna=True)` on the count side. Missing values continue to be reported separately by `MissingValueCount` (no double-counting).
- **`src/evidently/core/metric_types.py`**
  - Add `@validator("counts", "shares", pre=True)` on `ByLabelCountValue`, mirroring `ByLabelValue.convert_labels`.
  - Identity-check `pd.NA` in `convert_types` and normalize to `None`, consistent with `Label = Union[StrictBool, int, str, None]`. `np.nan` behavior is preserved (still falls through to pydantic's str-coercion path, matching the existing test contract).

## Tests

- **New** `tests/future/presets/test_dataset_stats_na.py` — end-to-end coverage matching the issue's exact reproduction on both `DataSummaryPreset` and `DataDriftPreset`.
- **Extended** `tests/future/metrics/test_unique_value_count.py` — cases for `pd.NA` in `string` / `object` / `Int64` dtypes.
- **Extended** `tests/future/test_metric_types.py` — parametrize cases covering `pd.NA` and `np.int64` keys on `ByLabelCountValue`.

### DIFF
--- a/src/evidently/core/metric_types.py
+++ b/src/evidently/core/metric_types.py
@@ -486,6 +486,12 @@ class ByLabelCountValue(MetricResult):
     count_display_name_template: str = "Missing label {label} count"
     share_display_name_template: str = "Missing label {label} share"
 
+    @validator("counts", "shares", pre=True)
+    def _convert_label_keys(cls, value):
+        if not isinstance(value, dict):
+            return value
+        return {convert_types(k): v for k, v in value.items()}
+
     def labels(self) -> List[Label]:
         return list(self.counts.keys())
 
@@ -551,6 +557,8 @@ except:  # noqa: E722
 
 
 def convert_types(val):
+    if val is pd.NA:
+        return None
     if isinstance(
         val,
         (
--- a/src/evidently/metrics/column_statistics.py
+++ b/src/evidently/metrics/column_statistics.py
@@ -869,9 +869,9 @@ class UniqueValueCountCalculation(ByLabelCountCalculation[UniqueValueCount]):
         return f"Unique Value Share: {self.metric.column} for label {label}"
 
     def _all_unique_values(self, current: Dataset, reference: Optional[Dataset]) -> set:
-        values = set(current.as_dataframe()[self.metric.column].unique())
+        values = set(current.as_dataframe()[self.metric.column].dropna().unique())
         if reference is not None:
-            values.update(reference.as_dataframe()[self.metric.column].unique())
+            values.update(reference.as_dataframe()[self.metric.column].dropna().unique())
         return values
 
     def _calculate_value(self, dataset: Dataset, values: set):
--- a/tests/future/metrics/test_unique_value_count.py
+++ b/tests/future/metrics/test_unique_value_count.py
@@ -1,5 +1,7 @@
+import numpy as np
 import pandas as pd
 
+from evidently import DataDefinition
 from evidently import Dataset
 from evidently import Report
 from evidently.core.metric_types import ByLabelCountValue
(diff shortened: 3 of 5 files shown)


ugbotueferhire:
hello @Liraim @DimaAmega hope you good I made a PR some days ago please review when you free thanks.

PR #1889, opened 2026-06-20
Fix SemanticSimilarity for sentence-transformers 5.4+

### LINKED ISSUE #1888: SemanticSimilarity fails with sentence-transformers > 5.3.0
`SemanticSimilarity` descriptor fails when installing Evidently with the latest available `sentence-transformers` version.

This looks like a dependency compatibility issue. The same example works if sentence-transformers is pinned to 5.3.0 or lower.

We may need to either:
- fix SemanticSimilarity compatibility with the latest sentence-transformers, or
- add an upper bound / tested version constraint for sentence-transformers.

## Reproducible example

Failing install:

```
!pip install evidently[llm]

import pandas as pd
from evidently import Report
from evidently import Dataset, DataDefinition
from evidently.descriptors import SemanticSimilarity

eval_data = pd.DataFrame([
    {
        "question": "Will my transaction go through if I don't have enough funds?",
        "context": "Overdraft protection allows transactions to be completed even if the account balance is insufficient. This service incurs a $35 fee per overdraft event unless linked to a backup account..",
        "answer": "Yes, your transaction will go through if overdraft protection is enabled, but a $35 fee will apply",
        "reference_answer": "Yes, with overdraft protection, your transaction will complete, but you will be charged $35."
    },
    {
        "question": "How do I block my card if it's lost?",
        "context": "To block a lost or stolen card, users should immediately navigate to the Cards section in the FinBot app, select the affected card, and tap 'Block card'. Blocking is instant and irreversible.",
        "answer": "Go to the Cards section, select your card, and tap 'Block card' to block it instantly.",
        "reference_answer": "Open the app, go to Cards, choose your card, and tap 'Block card'. Blocking is immediate."
    }
])

eval_df = Dataset.from_pandas(
    pd.DataFrame(eval_data))

eval_df.add_descriptors(descriptors=[
      SemanticSimilarity(columns=["answer", "reference_answer"], alias="Semantic Similarity"),
])
```
Error:

```
ValueError: Unsupported input type: Series. Expected one of: str, dict, PIL.Image.Image, np.ndarray, torch.Tensor
```

## Workaround

Pinning sentence-transformers fixes the issue:

```
pip install sentence-transformers==5.3.0
```

omkar-334:
Fixes #1888  

`SemanticSimilarity` was passing `pandas` `Series` objects directly to `SentenceTransformer.encode`. Newer `sentence-transformers` versions treat a `Series` as one unsupported input, so the descriptor failed at runtime.

This converts the filled columns to plain lists before encoding and adds a focused regression test for that boundary.


RK0297:
Nice fix! Just a heads up - _context_relevance.py (lines 43 & 45 in semantic_similarity_scoring) also passes a Series directly into model.encode(), so sentence-transformers 5.6.1+ will still fail there.

I have the complementary fix and regression test for _context_relevance.py ready and am opening a small PR to cover that second call site.

PR #1928, opened 2026-09-18
fix(g_test): normalize expected frequencies for different sample size…

### LINKED ISSUE #1914: G-test raises ValueError when reference and current datasets have different sample sizes
## Bug

The `g_test` statistical drift detector raises a `ValueError` when the
reference and current datasets have different sample sizes.

### Reproduction

```python
import pandas as pd

from evidently.legacy.calculations.stattests.g_stattest import _g_stat_test
from evidently.legacy.core import ColumnType

reference = pd.Series(["a"] * 500 + ["b"] * 500)
current = pd.Series(["a"] * 50 + ["b"] * 50)

_g_stat_test(reference, current, ColumnType.Categorical, 0.05)

### PR DESCRIPTION
**Pull request by shobhitagnihotri69**
### Summary
Fixes #1914.

When reference and current datasets have different sample sizes, `scipy.stats.power_divergence` (used by the G-test drift detector) raises `ValueError: f_obs and f_exp must have the same sum` because contingency table marginal frequencies were unscaled.

### Changes
1. Scaled `f_exp` proportionally by `sum(f_obs) / sum(f_exp)` before executing `power_divergence`.
2. Added unit regression test verifying G-test calculation executes cleanly and returns valid p-values when reference and current datasets have unequal sample sizes.

### CODE REVIEW

**Comment by feiiiiii5** [comment]
I ran this against `main` and against a hand-computed G statistic. The change is correct, and it matches an idiom that already exists in this repo.

**The fix reproduces the textbook expected frequencies.** With `reference = 500a/500b` and `current = 60a/40b` (n=100), the expected counts are `E = [50, 50]`, so

```
G = 2*(60*ln(60/50) + 40*ln(40/50)) = 4.0271,  df=1  ->  p = 0.044775
```

`_g_stat_test` with your patch returns `p = 0.04477477232924416` — the same number. Without it, that call raises `ValueError: For each axis slice, the sum of the observed frequencies must agree with the sum of expected frequencies`.

**This also removes an inconsistency between two sibling detectors.** `chisquare_stattest.py:42-44` already does exactly

```python
k_norm = current_data.shape[0] / reference_data.shape[0]
f_exp = [ref_feature_dict[key] * k_norm for key in keys]
```

so on the same unequal-size data the chi-square drift test answered while the G-test crashed:

| ref / cur | `_chi_stat_test` (main) | `_g_stat_test` (main) | `_g_stat_test` (yours) |
|---|---|---|---|
| 500/500 vs 50a50b | `(1.0, False)` | ValueError | `(1.0, False)` |
| 500/500 vs 60a40b | `(0.0455, True)` | ValueError | `(0.0448, True)` |
| 500/500 vs 50a49b1c | `(0.0, True)` | ValueError | `(0.0, True)` |

Copying that pair of lines is the right call for its own sake; if you are open to it, pulling the `keys`/`ref_feature_dict`/`f_exp`/`f_obs` block into one helper used by both files would be what keeps them from drifting apart again. Their two bodies are otherwise character-for-character identical, so the drift was easy to miss.

**Two things the tests do not yet cover.**

1. The new test uses a symmetric reference (`500/500`), where the correct scaling and a plain uniform split produce the *same* `f_exp`. I checked: with `f_exp = [n_cur/len(keys)] * len(keys)` — i.e. ignoring the reference distribution entirely — your test still passes, because both give `[50, 50]`. An asymmetric reference separates them: for `ref = 700a/300b`, `cur = 50a/50b`, correct scaling gives `p = 0.0000, drift=True`, while the uniform mistake gives `p = 1.0, drift=False`. Asserting an exact p-value on such a case (or on the 60/40 case above) would pin the scaling itself, not just the fact that it no longer raises.
2. `k_norm` uses full series lengths while `value_counts()` and `keys` exclude NaN, so when the two frames have *different* NaN ratios the sums still disagree and the call still raises. Measured with `ref = 250a/250b/500 NaN` and `cur = 50a/50b`: your patched `_g_stat_test` raises `ValueError`, and so does the existing `_chi_stat_test`. Identical NaN ratios on both sides work. This is pre-existing shared behaviour rather than something your change introduces, so it may belong in a separate issue — but it does mean #1914 is only fixed for columns without missing values, which is worth knowing before the issue is closed. Same family: an empty reference now reaches `k_norm` and raises `ZeroDivisionError` (also what `_chi_stat_test` already does; the G-test previously raised `ValueError` first).

I did not trace whether every upstream caller filters NaN before reaching these functions, so treat point 2 as a question for the maintainers rather than a defect claim.

Environment, so the numbers are reproducible: editable install of the repo in a Python 3.11 venv, calling `_g_stat_test` / `_chi_stat_test` directly; the "yours" column came from applying just your two lines to `g_stattest.py` on top of `main`.

*Disclosure: this comment was drafted with an AI coding assistant; every number above came from a command I ran locally, not from reading the diff.*

shobhitagnihotri69:
Thank you @feiiiiii5 for the thorough verification and mathematical breakdown! I've added `test_g_test_different_sample_sizes_asymmetric` in commit `880fa4c` to specifically test asymmetric reference distributions (700a/300b vs 50a/50b) to ensure the scaling factor pins expected frequencies rather than just a uniform split.

Issue #1930, opened 2026-09-18
TestSummaryInfo.has_all checks any_column: default TestSummary emits no aggregation metric, success_any alone crashes the report


feiiiiii5:
### What happens

`TestSummaryInfo.has_all` reports whether `any_column` is set, instead of `all_column`:

```python
@property
def has_all(self):
    """Check if 'all' column is configured.

    Returns:
    * `True` if `any_column` is set, `False` otherwise.
    """
    return self.any_column is not None
```

The summary line says "all", the body checks "any", and every sibling property (`has_any`, `has_count`, `has_rate`, `has_score`) returns its own column — so this reads as a copy-paste slip rather than intent.

`TestSummaryInfoPreset` gates the aggregation metric (`src/evidently/presets/special.py:34`) and the "All tests passed" widget (`:70`) on this flag, which produces two user-visible results.

### Reproduction

Offline, no API keys, on `main` @ `34771fe`, Python 3.11.15:

```python
import pandas as pd
from evidently.core.datasets import ColumnTest, Dataset, TestSummary, TestSummaryInfo
from evidently.tests import eq

def build(**kw):
    d = Dataset.from_pandas(pd.DataFrame({"a": [1, 1, 2]}))
    d.add_descriptor(ColumnTest("a", eq(1)))
    d.add_descriptor(TestSummary(alias="summary", **kw))
    return d

def info_of(d):
    return [i for i in d.data_definition.special_columns if isinstance(i, TestSummaryInfo)][0]
```

**1. The documented default silently produces no aggregation at all.** `TestSummary(success_all=True)` is the default (`success_all: bool = True`), so `all_column` is configured and `any_column` is not:

```
S1 default   all_column='summary' any_column=None has_all=False
S1 metric columns generated: []
```

`has_all` is `False`, so no aggregation metric is generated and the report has no "All tests passed" panel. Nothing raises and nothing warns.

**2. `success_any` on its own crashes the report.** With `all_column=None` and `any_column='summary'`, `has_all` is `True`, so the preset builds the "all" aggregation on a column that does not exist:

```
S2 any-only  all_column=None any_column='summary' has_all=True
S2 report RAISED ValidationError: 1 validation error for UniqueValueCount
```

Setting both flags works, which confirms the two columns are independent and only the flag is wired to the wrong field.

### Expected

`has_all` should follow `all_column`.

Fix + regression tests: https://github.com/evidentlyai/evidently/pull/1931

PR #1931, opened 2026-09-18
Fix TestSummaryInfo.has_all to check all_column

feiiiiii5:
## Problem

`TestSummaryInfo.has_all` returns whether `any_column` is set rather than `all_column` (`src/evidently/core/datasets.py:873-881`). `TestSummaryInfoPreset` gates both the aggregation metric and the "All tests passed" widget on that flag, so on `main`:

- the documented default `TestSummary()` (`success_all=True`) generates **no** aggregation metric — the panel is silently missing, with no error or warning;
- the documented `TestSummary(success_all=False, success_any=True)` raises `ValidationError: 1 validation error for UniqueValueCount` from `Report(metrics=[TextEvals()]).run(...)`, because the "all" aggregation is built on a column that was never configured.

Reproduction and the measured before/after output are in #1930.

## Change

One property now reads the field its own docstring names, matching its siblings:

```diff
     def has_all(self):
         """Check if 'all' column is configured.

         Returns:
-        * `True` if `any_column` is set, `False` otherwise.
+        * `True` if `all_column` is set, `False` otherwise.
         """
-        return self.any_column is not None
+        return self.all_column is not None
```

That is the whole production change — 2 lines. Everything else in this PR is tests.

## Verification

All commands ran in one venv (Python 3.11.15, editable install of the repo) against `main` @ `34771fe`, and each was repeated on the unpatched tree as a control with the identical command line.

- `python -m pytest tests/future/descriptors tests/future/presets tests/future/metrics tests/future/report tests/future/tests tests/future/test_data_definition.py tests/future/test_test_summary_columns.py -q`
  - base: `6 failed, 1610 passed, 6 skipped`
  - this branch: `1616 passed, 6 skipped`
  - the difference is exactly the six new cases that pin the defect; nothing else moves.
- The two scenarios from the issue, printed directly: on base `has_all=False` with an empty generated-metric list, and a `ValidationError` for the any-only report; on this branch `has_all=True` with `['summary']` generated, and the any-only report renders.
- Mutation checks, to confirm the tests are not vacuous: restoring the old `return` makes 6 of them fail; making `has_any` return `all_column` instead also makes 6 fail.
- `mypy`, run bare as `.github/workflows/main.yml:133` does: **142 errors in 34 files on base, and the identical 142 on this branch**, none of them in `core/datasets.py`. This change neither introduces nor fixes a type error; CI with the project's full dependency set is the authority on the pre-existing ones.
- Not covered locally: `tests/future/test_ui` and anything importing `evidently.ui.service.storage.sql` do not collect here because the `evidently[sql]` extra is absent, so I excluded those paths rather than run them. The full `python -m pytest` that CI runs was therefore not reproduced end to end.

## Scope

- One root cause: the mis-wired flag. I did not touch `presets/special.py`, widget layout, or the other `TestSummaryInfo` fields.
- The new tests assert on rendered widget titles obtained through `Report(...).run(...)`, which is the user-facing surface but does read the snapshot's private `_widgets`. Happy to rework them onto a public accessor if you prefer.
- Heads-up on overlap with #1847: that PR moves the `TestSummary` widget relative to the `TextEvals` metric widgets in `presets/dataset_stats.py`. Different file and concern, but if both land, widget ordering in the rendered report shifts, and the title assertions here could need a rebase.

---

Disclosure: this was investigated and written with an AI coding assistant. Every number above comes from a command run against `main` @ `34771fe`, and each was repeated on the unpatched tree with the identical command line so the delta is independently checkable. I am happy to answer for the change itself.


feiiiiii5:
Note: #1747 (open since Nov 2025, no activity for ~10 months) proposes the same one-line change to `has_all`. This PR was opened without that PR on our radar; the difference here is the full measured reproduction (#1930), the two new tests, and verification of the emitted metrics/widgets on both sides. Happy to defer to the earlier PR if maintainers prefer — whichever route is chosen, the stale one could be closed so this doesn't stay ambiguous.

